PR #7, opened 2019-07-04
Feature/html validation

katedee:
## This PR covers the following trello tickets:
- [Add form validation to HTML prototype](https://trello.com/c/8D5rgaWX/108-add-form-validation-to-html-prototype)
- [Add error summary element to HTML prototype](https://trello.com/c/KjG2Nq44/109-add-error-summary-element-to-html-prototype)

Just enter nothing on the page asking for your code to see.

## What it do
I've added this html validation setup to the login page, which may seem overkill, but will serve as reference for how to set this up.

Schemas for validation are contained in one file right now, and we import them where appropriate.

Essentially we check the schema in our route, nab the validation results with `validationResult`, and if there are errors, render the page with the errors using the `errorArray2ErrorObject` function I 100% stole from @pcraig3 

When errors are present, we include `errorList`, which list out the errors on the page, and links to the appropriate input. This error list div has a `role=alert` for some [https://www.w3.org/TR/WCAG20-TECHS/ARIA19.html](error accessibility standards)

Additionally, when there are errors, we get an inline error, and the input gets a `aria-describedby` which links up the appropriate message in the error list. The input also gets a red border as visual indication beyond the error below.

## How to do
I've made some assumptions here, and based on review things may change, but for implementing this, the main points to look at are:

- write your schema in `formSchemas.js`
- check your schema in your controller file when making request (`login.controller.js` ~line 15)
- store the results (`login.controller.js` ~line 34) -- this example is a bit different as at this point we know there are errors, you might run an extra `if(!errors.isEmpty())`
- render appropriate page with errors being passed to `errorObject2ErrorArray()` (`login.controller.js` ~line 36)

and then for your inputs, make sure you're checking if it should have the class `has-error`, if it should have an `aria-describedby` and a trailing error message in a span:
```
input#code(class={['has-error']: errors && errors.code} name='code', type='text', value=data.code, autocomplete='off' aria-describedby=(errors && errors.code ? 'code_error' : false))
          if errors
            span.validation-message #{errors.code.msg}
```

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,5 @@ typings/
 # CSS files are compiled
 public/styles.css
 public/styles.css*
+
+**/**/.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -68,4 +68,4 @@ typings/
 public/styles.css
 public/styles.css*
 
-**/**/.DS_Store
+.DS_Store
--- a/formSchemas.js
+++ b/formSchemas.js
@@ -4,6 +4,9 @@ const loginSchema = {
         errorMessage: 'Must be 8 characters',
         options: { min: 8, max: 8 }
       },
+      isAlphanumeric: {
+          errorMessage: 'Code can only contain letters and numbers'
+      }
     },
   }
 
--- a/formSchemas.js
+++ b/formSchemas.js
@@ -1,6 +1,5 @@
 const loginSchema = {
     code: {
-      in: ['body'],
       isLength: {
         errorMessage: 'Must be 8 characters',
         options: { min: 8 }
--- a/formSchemas.js
+++ b/formSchemas.js
@@ -1,0 +1,13 @@
+const loginSchema = {
+    code: {
+      in: ['body'],
+      isLength: {
+        errorMessage: 'Must be 8 characters',
+        options: { min: 8 }
+      },
+    },
+  }
+
+  module.exports = {
+    loginSchema
+  }
--- a/formSchemas.js
+++ b/formSchemas.js
@@ -2,7 +2,7 @@ const loginSchema = {
     code: {
       isLength: {
         errorMessage: 'Must be 8 characters',
-        options: { min: 8 }
+        options: { min: 8, max: 8 }
       },
     },
   }
--- a/public/scss/_error-list.scss
+++ b/public/scss/_error-list.scss
@@ -1,0 +1,13 @@
+.error-list-wrapper {
+    padding: 15px;
+    background: #f3e9e8;
+    border-left: 4px solid $color-red;
+
+    h3 {
+        margin-top: 10px;
+    }
+}
+
+.validation-message {
+  color: $color-red;
+}
--- a/public/scss/_error-list.scss
+++ b/public/scss/_error-list.scss
@@ -1,10 +1,10 @@
 .error-list {
-    padding: 15px;
-    background: #f3e9e8;
+    padding: $space-sm;
+    background: $color-red-light;
     border-left: 4px solid $color-red;
 
     &__header {
-        margin-top: 10px;
+        margin-top: $space-xs;
     }
 }
 
--- a/public/scss/_error-list.scss
+++ b/public/scss/_error-list.scss
@@ -1,9 +1,9 @@
-.error-list-wrapper {
+.error-list {
     padding: 15px;
     background: #f3e9e8;
     border-left: 4px solid $color-red;
 
-    h3 {
+    &__header {
         margin-top: 10px;
     }
 }
--- a/routes/login/login.controller.js
+++ b/routes/login/login.controller.js
@@ -1,10 +1,6 @@
 const { validationResult, checkSchema } = require('express-validator')
-const {
-  errorArray2ErrorObject
-} = require('./../../utils.js')
-const {
-  loginSchema
-} = require('./../../formSchemas.js')
+const { errorArray2ErrorObject } = require('./../../utils.js')
+const { loginSchema } = require('./../../formSchemas.js')
 
 module.exports = function(app) {
   // redirect from "/login" → "/login/accessCode"
@@ -27,11 +23,11 @@ const postCode = (req, res) => {
   let accessCode = req.body.code || null
   req.session = accessCode ? { code: accessCode } : null
 
-  if (accessCode && redirect) {
-    return res.redirect(redirect)
-  }
-
   const errors = validationResult(req)
 
-  res.status(422).render('login/code', { title: 'Enter access code', data: req.session || {}, errors: errorArray2ErrorObject(errors) })
+  if(!errors.isEmpty()) {
+    res.status(422).render('login/code', { title: 'Enter access code', data: req.session || {}, errors: errorArray2ErrorObject(errors) })
+  } else if (accessCode && redirect) {
+    return res.redirect(redirect)
+  }
 }
--- a/routes/login/login.controller.js
+++ b/routes/login/login.controller.js
@@ -1,10 +1,18 @@
+const { validationResult, checkSchema } = require('express-validator')
+const {
+  errorArray2ErrorObject
+} = require('./../../utils.js')
+const {
+  loginSchema
+} = require('./../../formSchemas.js')
+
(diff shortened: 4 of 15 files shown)


utils.js
```diff
@@ -0,0 +1,29 @@
+const pickEditSchema = (req, res, next) => {
+    const question = require(`./questions/${req.params.id}.js`)
```
sboisvert: Could someone break out of the folder by passing an ID that has ../../ in it? Potentially loading a file we don't expect them to load?

utils.js
```diff
@@ -0,0 +1,29 @@
...
+    { param: 'number', msg: 'Cannot be empty', ... }
+  ]
+  to
+  {
+    name: { param: 'name', msg: 'Cannot be empty', ... },
+    number: { param: 'number', msg: 'Cannot be empty', ... }
+  }
+*/
+const errorArray2ErrorObject = (errors = []) => {
+    return errors.array({ onlyFirstError: true }).reduce((map, obj) => {
```
sboisvert: Does `onlyFirstError: true` mean that if I mess up 2 fields I only get shown the first error? If so, is this what users expect? (Personally, I really don't know)

formSchemas.js
```diff
@@ -0,0 +1,13 @@
+const loginSchema = {
```
sboisvert: Will we create a new object for each "page" or for each "section"? (Section being login, personal info, tax info, tax credits vs pages being "enter your access code, enter your SIN, DoB, etc)

.gitignore
```diff
@@ -66,4 +66,6 @@ typings/
 
 # CSS files are compiled
 public/styles.css
-public/styles.css*
\ No newline at end of file
+public/styles.css*
+
+**/**/.DS_Store
```
pcraig3: I think if you just do `.DS_Store`, this will still work?

formSchemas.js
pcraig3: I think we should 'section' them at this point.

So we would have 

```
const loginSchema = {
  code: {...},
  sin: {...},
  dob: {...}
}
````

File: formSchemas.js [hunk 5]
```diff
@@ -0,0 +1,13 @@
...
+      isLength: {
+        errorMessage: 'Must be 8 characters',
+        options: { min: 8 }
+      },
+    },
+  }
+
+  module.exports = {
+    loginSchema
+  }
```
pcraig3: Small point, but I'm going to figure out how to get trailing newlines to be automatically inserted as they were on past projects.

public/scss/_error-list.scss
```diff
@@ -0,0 +1,13 @@
+.error-list {
+    padding: 15px;
+    background: #f3e9e8;
+    border-left: 4px solid $color-red;
+
+    &__header {
+        margin-top: 10px;
+    }
```
pcraig3: I've added some CSS variables for spacing here: https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/public/scss/_variables.scss#L11

For `height`s and `width`s, it doesn't really matter, but I think for `padding`s and `margin`s, we should try and use them to be consistent.

Also, let's put in your new red background as a new color.

routes/login/login.spec.js
```diff
@@ -33,6 +33,28 @@ describe('Test /login responses', () => {
...
+    test('it renders an inline error for /login/code with appropriate describedby', async () => {
+      const response = await request(app)
+        .post('/login/code')
+        .send({ redirect: '/' })
+      const $ = cheerio.load(response.text)
+      expect($('.validation-message').text()).toEqual('Must be 8 characters')
+      expect($('#code').attr('aria-describedby')).toEqual('code_error')
+    })
+  })
+
```
pcraig3: Great tests! 

If the text changes later, the tests will break and then we can re-evaluate whether to continue to check for the while string or a smaller subset, but for now this looks great! 💯 💯

utils.js
pcraig3: It's more for cases where multiple validations fail for one field.

Like if you wanted to say:
- sin must be not empty
- sin must be 9 characters
- sin can only be numbers

If I entered "SIN", then i would get back 2 errors: "9 chars" and "only numbers". Asking for just the first error message would only display "9 characters" and then that's what would show up on the form.

I think for our current purposes, this is fine.

views/_includes/errorList.pug
```diff
@@ -0,0 +1,6 @@
+div.error-list(role='alert')
+    h3.error-list__header #{__('Please correct the errors on the page')}
+    ol#formErrors.error-list__list
+        each error in errors
+            li.error-list__list-item
+                a(class='error-list__link' id=error.param+'_error' href='#'+error.param) #{error.msg}
```
pcraig3: I've used BEM in the past too! I haven't been doing anything specific in this repo so far but if you want to continue with BEM CSS conventions, I'm game 👍

views/base.pug
```diff
@@ -27,6 +27,9 @@ html(lang='en')
               img(src='/img/sig-blk-en.svg', alt='Government of Canada')
 
       main
+        if errors 
+          include _includes/errorList.pug
+
```
pcraig3: Awesome! v clean.

utils.js
pcraig3: Yeah, good question

pcraig3: Although, the slashes would be hard to pass in considering how urls work

views/login/code.pug
```diff
@@ -11,7 +11,9 @@ block content
         div
           label(for='code') Personal access code
           span.pure-form-message For Example: A23XGY12
-          input#code(name='code', type='text', value=data.code, autocomplete='off')
+          input#code(class={['has-error']: errors && errors.code} name='code', type='text', value=data.code, autocomplete='off' aria-describedby=(errors && errors.code ? 'code_error' : false))
```
pcraig3: What does this syntax mean? 
ie, `class={['has-error']: errors && errors.code}`?

utils.js
katedee: oh shoot that got left over and shouldn't be there! thanks for catching!

formSchemas.js
katedee: I imagine for each form, which happens to get broken down into page, but this is up for discussion! As I don't have more forms to work with right now--undecided! but good question

views/login/code.pug
katedee: so this is just conditionally checking wether the class should be added, based on if the error for that specific input is present (but we have to check if errors are present at all first).

[pug's got the write up on it here](https://pugjs.org/language/attributes.html) (under class attributes)

formSchemas.js
pcraig3: yeah, not a big deal for this PR

routes/login/login.controller.js
```diff
@@ -19,9 +23,11 @@ const postCode = (req, res) => {
...
-  if (accessCode && redirect) {
+  const errors = validationResult(req)
+
+  if(!errors.isEmpty()) {
+    res.status(422).render('login/code', { title: 'Enter access code', data: req.session || {}, errors: errorArray2ErrorObject(errors) })
+  } else if (accessCode && redirect) {
     return res.redirect(redirect)
   }
-
-  res.status(422).render('login/code', { title: 'Enter access code', data: req.session || {} })
```
pcraig3: Hmm, this method doesn't return anything by default. 

I think it should be changed so that there's one `if`, and then falls through to some other behaviour.

But for now it's okay: we can validate `redirect` is there in some middleware and then we don't have to run that check in here.

views/login/code.pug
pcraig3: Ah, I see, the array syntax plus the backwards ternary logic threw me.

formSchemas.js
```diff
@@ -0,0 +1,12 @@
+const loginSchema = {
+    code: {
+      isLength: {
+        errorMessage: 'Must be 8 characters',
+        options: { min: 8 }
```
pcraig3: A-ha! This lets codes with 9 characters through still.

pcraig3:
Awesome! Let's merge 👍 👍 👍

PR #12, opened 2019-07-08
Login with access code + put user data into session

pcraig3:
Previously, we were accepting an access code as long as it was formatted correctly and then displaying it on the next page.

Since our app works by accepting a user's access code and then pulling in a bunch of user data, I've changed the access code validation to only work if a matching access code is found (currently there is only one: `QWER1234`), and then user data is inserted into the session. Just to do something with it, I inserted the user's name onto the following 'success' page.  I also updated the hint on the page to give the correct access code, and rewrote a few of the tests.

I've also changed the error messages into dot notation so that we can translate them when the time comes.

## Screenshot

Screen Shot 2019-07-08 at 9 16 57 AM

### DIFF
--- a/api/index.spec.js
+++ b/api/index.spec.js
@@ -11,3 +11,9 @@ test('returns null with a nonexistent login.code', () => {
   const user = API.getUser('H3LLY34H')
   expect(user).toBe(null)
 })
+
+test('returns matching login code as an array', () => {
+  const codes = API.getMatches()
+  expect(codes).toBeInstanceOf(Array)
+  expect(codes.length).toBe(1)
+})
--- a/api/index.spec.js
+++ b/api/index.spec.js
@@ -1,7 +1,7 @@
 const API = require('./index')
 
 test('returns expected user with correct login.code', () => {
-  const user = API.getUser('ABCD1234')
+  const user = API.getUser('QWER1234')
   expect(user).not.toBe(null)
   expect(user.personal.firstName).toBe('Gabrielle')
   expect(user.personal.lastName).toBe('Roy')
--- a/locales/en.json
+++ b/locales/en.json
@@ -1,6 +1,8 @@
 {
-	"This service is to claim tax benefits online.": "This service is to claim tax benefits online.",
-	"Claim tax benefits": "Claim tax benefits",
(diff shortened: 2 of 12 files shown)


routes/login/login.spec.js
```diff
@@ -41,7 +41,7 @@ describe('Test /login responses', () => {
       const $ = cheerio.load(response.text)
       expect($('.error-list__header').text()).toEqual('Please correct the errors on the page')
       expect($('.error-list__list').children()).toHaveLength(1)
-      expect($('.validation-message').text()).toEqual('Must be 8 characters')
+      expect($('.validation-message').text()).toEqual('Access code must be 8 characters')
```
katedee: 👏

api/index.spec.js
```diff
@@ -1,7 +1,7 @@
 const API = require('./index')
 
 test('returns expected user with correct login.code', () => {
-  const user = API.getUser('ABCD1234')
+  const user = API.getUser('QWER1234')
   expect(user).not.toBe(null)
   expect(user.firstName).toBe('Gabrielle')
```
sboisvert: J'aime la référence

locales/en.json
```diff
@@ -1,5 +1,8 @@
 {
-	"This service is to claim tax benefits online.": "This service is to claim tax benefits online.",
-	"Claim tax benefits": "Claim tax benefits",
-	"Please correct the errors on the page": "Please correct the errors on the page"
-}
\ No newline at end of file
+  "This service is to claim tax benefits online.": "This service is to claim tax benefits online.",
+  "Claim tax benefits": "Claim tax benefits",
+  "Please correct the errors on the page": "Please correct the errors on the page",
+  "errors.login.length": "Access code must be 8 characters",
```
sboisvert: Why are we having some strings with identifiers and others without?

pcraig3: The validation messages were being put directly in the schemas, which is a third place, so I didn't want to keep them there.

The other two places are:

1. directly in the template
2. in the i18n locales json files

Since error messages are conceptually related, they're an easy candidate to pick off, so that we group them together in the i18n files and update them all at once down the road.

PR #14, opened 2019-07-08
(C)Lickable buttons

pcraig3:
> “We made the buttons on the screen look so good you’ll want to lick them.”
-- Steve Jobs

This is a small PR, just focused on the buttons.

In here, we make the buttons actually click. Looks way better.

Moved the button-related CSS into a new 'components' directory that we can use to put self-contained, reusable bits of interface.

Source:
- https://azumbrunnen.me/blog/the-art-of-the-button/
- https://design-system.service.gov.uk/components/button/

## gifs

| before | after |
|--------|-------|
|   DOESN'T MOVE + focus is weird     |  👨‍🎨👌    |
|  button     |    button2  |

### DIFF
--- a/public/scss/_forms.scss
+++ b/public/scss/_forms.scss
@@ -1,18 +1,3 @@
-%button-styles {
-  line-height: 1.1875;
-  display: inline-block;
-  position: relative;
-  margin-top: 10px;
-  padding: 7px 10px;
-  border-radius: 0;
-  color: $color-white;
-  background-color: $color-blue-light;
-  border: 2px solid $color-blue-light;
-  text-align: center;
-  cursor: pointer;
-  text-decoration: none;
-}
-
 form.pure-form.pure-form-stacked {
   margin-top: $space-xl;
   max-width: 300px;
@@ -51,8 +36,7 @@ form.pure-form.pure-form-stacked {
     font-weight: 700;
   }
 
-  input,
-  button {
+  input {
     font-family: sans-serif;
     font-size: 1em;
     font-weight: 400;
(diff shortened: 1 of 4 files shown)


public/scss/_forms.scss
```diff
@@ -78,11 +62,6 @@ form.pure-form.pure-form-stacked {
     }
   }
 
-  button {
-    @extend %button-styles;
-    font-family: 'Noto Sans', sans-serif;
-  }
-
   .buttons {
     display: flex;
```
katedee: 👌 but what kinda browser support we looking at? Might need some fallback styles if we're doing .
IE 10 needs the `ms-` prefix, IE 11 we'll just have to double check because it can get weird, and IE 9 support is 😬

public/scss/components/_buttons.scss
```diff
@@ -0,0 +1,71 @@
+%button-styles {
```
katedee: 🙌

views/start/index.pug
```diff
@@ -10,4 +10,4 @@ block content
     li Personal Code
     li Social Insurance Number (SIN)
     li Income Information
-  a.button-link(href='/login/code' role='button' draggable='false') Start now →
+  a.button-link(href='/login/code' role='button' draggable='false') Start now <span aria-hidden="true">→</span>
```
katedee: nice 👌

public/scss/_forms.scss
pcraig3: Currently, we haven't discussed browser support. 

My opinion is that we want IE11 for sure, and maybe IE10.

We have access to [browserstack](https://www.browserstack.com/), so we can test with different browsers, but I think we should leave this until next sprint.

pcraig3:
@katedee good question about browser support. 

We want to hit IE11 for sure, and for IE10, I think let's see how bad it gets. But for this sprint, we don't care yet. 

I've added [a card](https://trello.com/c/b3WDWpKL/119-wip-polyfills-for-ie11-maybe-10) and we can pick it up next sprint.

PR #17, opened 2019-07-08
Add error validation for SIN

sboisvert:
Uses a regex to remove spaces, dashes and underscores then checks that it's 9 characters and then checks that it's an int.

### DIFF
--- a/formSchemas.js
+++ b/formSchemas.js
@@ -29,7 +29,7 @@ const sinSchema = {
       },
     },
     isLength: {
-      errorMessage: 'errors.login.length',
+      errorMessage: 'errors.login.lengthSIN',
       options: { min: 9, max: 9 },
     },
     isInt: {
--- a/formSchemas.js
+++ b/formSchemas.js
@@ -18,9 +18,27 @@ const loginSchema = {
       options: [API.getMatches()],
       errorMessage: 'errors.login.code',
     },
+  }
+}
+const sinSchema = {
+  sin: {
+    customSanitizer: {
+      options: value => {
+        //We want to remove any spaces, dash or underscores
+        return value ? value.replace(/[ \-_]*/g, '') : value
+      },
+    },
+    isLength: {
+      errorMessage: 'errors.login.length',
+      options: { min: 9, max: 9 },
+    },
+    isInt: {
+      errorMessage: 'errors.login.numericSIN',
+    },
   },
 }
 
 module.exports = {
   loginSchema,
+  sinSchema,
 }
--- a/routes/login/login.controller.js
+++ b/routes/login/login.controller.js
@@ -1,9 +1,9 @@
 const { validationResult, checkSchema } = require('express-validator')
 const { errorArray2ErrorObject } = require('./../../utils.js')
-const { loginSchema } = require('./../../formSchemas.js')
+const { loginSchema, sinSchema } = require('./../../formSchemas.js')
 const API = require('../../api')
 
-module.exports = function(app) {
+module.exports = function (app) {
   // redirect from "/login" → "/login/code"
   app.get('/login', (req, res) => res.redirect('/login/code'))
   app.get('/login/code', (req, res) => res.render('login/code', { data: req.session || {} }))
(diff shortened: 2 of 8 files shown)


formSchemas.js
```diff
@@ -18,9 +18,27 @@ const loginSchema = {
...
       errorMessage: 'errors.login.code',
     },
+  }
+}
+const sinSchema = {
+  sin: {
+    customSanitizer: {
+      options: value => {
+        //We want to remove any spaces, dash or underscores
+        return value ? value.replace(/[ \-_]*/g, '') : value
```
pcraig3: Is there a way to pass non-string values back as post parameters? 

I would kind of think we're okay, but maybe see what an empty value or a single digit number end up as (probably a good test).

formSchemas.js
```diff
@@ -18,9 +18,27 @@ const loginSchema = {
...
+      },
+    },
+    isLength: {
+      errorMessage: 'errors.login.length',
```
pcraig3: Should this be SIN specific, or is this just a default error message? (either is fine right now, in my opinion)

pcraig3: Just noticed the error message for this is only for access codes: `Access code must be 8 characters`

pcraig3:
Also, something @nmakuch and I were doing before was putting in screenshots or gifs if there was a new page or some new behaviour. (like this one: https://github.com/cds-snc/cra-claim-tax-benefits/pull/16) 

It's maybe less needed now that we have the github PRs, but it's still kind of nice if you have an easy way to snap a photo.

routes/login/login.controller.js
```diff
@@ -57,6 +57,18 @@ const postSIN = (req, res) => {
...
   req.session.sin = sin
 
+  const errors = validationResult(req)
+
+  if (!errors.isEmpty()) {
+    // clear session
+    req.session = null
+
+    return res.status(422).render('login/sin', {
+      data: { code: req.body.code } || {},
```
pcraig3: To preserve the value that's passed in by the user (instead of clearing it), let's change this to `sin: req.body.sin` instead.

routes/login/login.spec.js
```diff
@@ -87,6 +89,78 @@ describe('Test /login responses', () => {
...
+      test('it renders an inline error for /login/code with appropriate describedby', async () => {
+        const response = await request(app)
+          .post('/login/code')
+          .send({ redirect: '/' })
+        const $ = cheerio.load(response.text)
+        expect($('.validation-message').text()).toEqual('Access code must be 8 characters')
+        expect($('#code').attr('aria-describedby')).toEqual('code_error')
+      })
+    })
+  */
```
pcraig3: yeah, let's delete this.

routes/login/login.spec.js
```diff
@@ -87,6 +89,78 @@ describe('Test /login responses', () => {
...
+  })
+
+  test('it does not allow non-numeric characters', async () => {
+    const response = await request(app)
+      .post('/login/code')
+      .send({ code: 'A23X456@1', redirect: '/' })
+    expect(response.statusCode).toBe(422)
+  })
+
+  //todo: validate that SIN is valid and matches the user's
```
pcraig3: I don't think 🤞 we need more unit tests for this. Instead, let's near the end write one/two functional tests that go through the app with some user fixture.

views/login/sin.pug
```diff
@@ -15,16 +15,16 @@ block content
     div
       label(for='sin') Social Insurance Number
       span.pure-form-message For Example: 123 456 789
-      input.w-3-4#sin(class={['has-error']: errors && errors.code} name='sin', type='text', value=data.code, autocomplete='off' aria-describedby=(errors && errors.code ? 'code_error' : false))
+      input.w-3-4#sin(class={['has-error']: errors && errors.code} name='sin', type='text', value=data.code, autocomplete='off' aria-describedby=(errors && errors.sin ? 'sin_error' : false))
```
pcraig3: - `errors.code` should be `errors.sin`
- `value=data.code` should be `value=data.sin`

PR #18, opened 2019-07-08
Feature/marital status page

katedee:
### Confirm + Edit Marital Status
I went with a pattern of `personal/[personalDetail]/edit` for the edit pages (so `personal/maritalStatus/edit`). Happy to change it if the team thinks we should structure it otherwise!

It doesn't currently pull in from the "API", but it's set up to receive those values.

Added a css "benefits-form" class, because I found I was fighting pure-form for the radio-buttons, but let me know if this is no good!

Edit:
- added focus styles to radio buttons
- added basic validation to the marital status page (this will still need to be update once we have the full user flow)
- moved validateRedirect to utils, to use as middleware

### More Edits:
- Add no-console rule to eslint (Disabled that rule for bin/www since we want console.logs there)
- Updated schema for maritalStatus
- Updated error code for maritalStatus to not say “below”
- Tidied up postSin
- Added some passing of maritalStatus in maritalStatus routes
- Added a few more tests for maritalStatus pages
- Re-ordered radio items

image

image

### DIFF
--- a/formSchemas.js
+++ b/formSchemas.js
@@ -38,11 +38,12 @@ const sinSchema = {
   },
 }
 
+//TODO: We'll want to store this array of marital options somewhere. This is temporary. I'll also want to use that later to create the radio buttons dynamically, to avoid having to update multiple files
 const maritalStatusSchema = {
   maritalStatus: {
-    isLength: {
+    isIn: {
       errorMessage: 'errors.maritalStatus.maritalStatus',
-      options: { min: 1 }
+      options: [['Married','Widowed','Divorced','Separated','Single']]
     }
   }
 }
--- a/formSchemas.js
+++ b/formSchemas.js
@@ -21,6 +21,16 @@ const loginSchema = {
   },
 }
 
+const maritalStatusSchema = {
+  maritalStatus: {
+    isLength: {
+      errorMessage: 'errors.maritalStatus.maritalStatus',
+      options: { min: 1 }
+    }
+  }
+}
+
 module.exports = {
   loginSchema,
+  maritalStatusSchema
 }
--- a/public/scss/_forms.scss
+++ b/public/scss/_forms.scss
@@ -85,13 +85,23 @@ form.pure-form.pure-form-stacked {
   }
 }
 
+.fieldset {
+  border: 0;
+  padding: 0;
+
+  &__legend {
+    font-weight: 600;
+    font-size: 19px;
+    font-size: 1.1875rem;
+  }
+}
+
 .radios {
   $radioClass: &;
 
   &__item {
-    font-weight: 400;
-    font-size: 16px;
-    font-size: 1rem;
+    font-size: 19px;
+    font-size: 1.1875rem;
     line-height: 1.25;
     display: block;
     position: relative;
--- a/public/scss/_forms.scss
+++ b/public/scss/_forms.scss
@@ -84,3 +84,71 @@ form.pure-form.pure-form-stacked {
     }
   }
 }
+
+.radios {
+  $radioClass: &;
+
+  &__item {
+    font-weight: 400;
+    font-size: 16px;
+    font-size: 1rem;
+    line-height: 1.25;
+    display: block;
+    position: relative;
+    min-height: 40px;
+    margin-bottom: 10px;
+    padding-left: 40px;
+  }
+
+  &__input {
+    cursor: pointer;
+    position: absolute;
+    z-index: 1;
+    top: -2px;
+    left: -2px;
+    width: 44px;
+    height: 44px;
+    margin: 0;
+    opacity: 0; 
+    
+    &:checked + #{$radioClass}__label::after {
+      opacity: 1;
+    } 
+  }
+
+  &__label {
+    display: inline-block;
+    margin-bottom: 0;
+    padding: 8px 15px 5px;
+    cursor: pointer;
+    -ms-touch-action: manipulation;
+    touch-action: manipulation;
+
+    &::before {
+      content: "";
+      -webkit-box-sizing: border-box;
+      box-sizing: border-box;
+      position: absolute;
+      top: 0;
+      left: 0;
+      width: 40px;
+      height: 40px;
+      border: 2px solid currentColor;
+      border-radius: 50%;
+      background: transparent;
+    }
+
+    &::after {
+      content: "";
+      position: absolute;
+      top: 10px;
+      left: 10px;
+      width: 0;
+      height: 0;
+      border: 10px solid currentColor;
+      border-radius: 50%;
+      opacity: 0;
+      background: currentColor;
+    }
+  }
+}
--- a/public/scss/_forms.scss
+++ b/public/scss/_forms.scss
@@ -35,7 +35,7 @@ form.cra-form {
     font-size: inherit;
   }
 
-  label {
+  label:not(.radios__label) {
     font-weight: 700;
     display: block;
   }
@@ -131,6 +131,11 @@ form.cra-form {
     &:checked + #{$radioClass}__label::after {
       opacity: 1;
     } 
+
+    &:focus + #{$radioClass}__label::before {
+      -webkit-box-shadow: 0 0 0 3px $color-yellow;
+      box-shadow: 0 0 0 3px $color-yellow;
+    }
   }
 
   &__label {
--- a/public/scss/_forms.scss
+++ b/public/scss/_forms.scss
@@ -93,7 +93,13 @@ form.pure-form.pure-form-stacked {
     font-weight: 600;
     font-size: 19px;
     font-size: 1.1875rem;
+    border-bottom: 0;
   }
+}
+
+// add this class because I was fighting pure form too much
+.benefits-form {
+  margin-bottom: $space-xl;
 }
 
 .radios {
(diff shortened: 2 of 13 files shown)


public/scss/_forms.scss
```diff
@@ -84,3 +84,87 @@ form.pure-form.pure-form-stacked {
...
+  }
+}
+
+// add this class because I was fighting pure form too much
+.benefits-form {
+  margin-bottom: $space-xl;
+}
+
+.radios {
+  $radioClass: &;
```
pcraig3: this is neat  👀

views/personal/maritalStatus.pug
```diff
@@ -0,0 +1,34 @@
...
+      thead
+        tr
+          th Marital Status
+      tbody
+        tr
+          td
+            if data && data.personal.maritalStatus
+              div #{data.personal.maritalStatus}
+            else
+              div Married
```
pcraig3: Cool, we can remove this `if` check later but good idea.

views/personal/maritalStatus-edit.pug
```diff
@@ -0,0 +1,45 @@
...
+                    label.radios__label(for="maritalStatusWidowed") Widowed
+                .radios__item
+                    input.radios__input(id="maritalStatusDivorced" name="maritalStatus" type="radio" value="Divorced" checked=(data && data.personal.maritalStatus.toLowerCase()=='divorced'))
+                    label.radios__label(for="maritalStatusDivorced") Divorced
+                .radios__item
+                    input.radios__input(id="maritalStatusSeparated" name="maritalStatus" type="radio" value="Separated" checked=(data && data.personal.maritalStatus.toLowerCase()=='separated'))
+                    label.radios__label(for="maritalStatusSeparated") Separated
+                .radios__item
+                    input.radios__input(id="maritalStatusSingle" name="maritalStatus" type="radio" value="Single" checked=(data && data.personal.maritalStatus.toLowerCase()=='single'))
+                    label.radios__label(for="maritalStatusSingle") Single
```
pcraig3: Maybe eventually we should make these an `/includes` as well, but for now this looks good 👍

views/personal/maritalStatus-edit.pug
```diff
@@ -0,0 +1,45 @@
...
+  -var title = 'Change your Marital Status'
+
+block content
+
+  h1 #{title}
+
+  div
+    p Please choose the correct status from the options listed below.
+
+    checked=(data && data.personal.maritalStatus.toLowerCase()=='married')
```
pcraig3: I think the `-var ` prefix is necessary here if you're using it — which I'm not sure that you are. 🤔 

Currently, this creates an empty `<checked></checked>` tag under the paragraph.

routes/personal/personal.spec.js
```diff
@@ -1,9 +1,43 @@
...
+      const $ = cheerio.load(response.text)
+      expect($('#maritalStatusMarried').attr('checked')).toEqual('checked')
+    })
+
+    test('it checks the stored marital status by default for /personal/maritalStatus/edit path', async () => {
+      const user = API.getUser('QWER1234') 
+      const response = await request(app).get('/personal/maritalStatus/edit', {data: user})
+      const $ = cheerio.load(response.text)
+      expect($('input[name=maritalStatus]:checked').val().toLowerCase).toEqual(user.personal.maritalStatus.toLowerCase)
+    })
```
pcraig3: Looks good! 👍

public/scss/_forms.scss
```diff
@@ -84,3 +84,87 @@ form.pure-form.pure-form-stacked {
...
+      left: 10px;
+      width: 0;
+      height: 0;
+      border: 10px solid currentColor;
+      border-radius: 50%;
+      opacity: 0;
+      background: currentColor;
+    }
+  }
+}
```
pcraig3: Looks fairly clean!

The only thing we're missing is the focus styling. I think [this](https://github.com/cds-snc/cra-alpha/blob/master/src/components/forms/MultipleChoice.js#L104) should work. 

What we're looking for is that the currently focused checkbox has a yellow ring around it.

views/personal/maritalStatus-edit.pug
```diff
@@ -0,0 +1,45 @@
...
+                    input.radios__input(id="maritalStatusSingle" name="maritalStatus" type="radio" value="Single" checked=(data && data.personal.maritalStatus.toLowerCase()=='single'))
+                    label.radios__label(for="maritalStatusSingle") Single
+
+  div
+    details.
+      <summary><span>Help with editing marital status</span></summary>
+      <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.</p>
+
+  .buttons-row
+    a.button-link(href='/personal/maritalStatus') Change Marital Status
```
pcraig3: You could do it as a later PR if you want, but we want this form to actually submit stuff. 

To do that, we should pull the `div > details` and the buttons into the form itself, and make this a real-life button instead of an `a` tag.

That would also include adding validation as well, so we can do it as a separate PR if you'd rather.

views/personal/maritalStatus-edit.pug
katedee: that should be there and not sure how it go there 🤔 
good catch! will remove

views/personal/maritalStatus-edit.pug
katedee: It shouldn't be too much work, so I'll just add it on to this PR

routes/personal/personal.controller.js
```diff
@@ -1,3 +1,31 @@
...
+
+  const errors = validationResult(req)
+
+  if (!errors.isEmpty()) {
+    return res.status(422).render('personal/maritalStatus-edit', {
+      errors: errorArray2ErrorObject(errors),
+    })
+  }
+
+  console.log(req.body.maritalStatus)
```
pcraig3: Let's get rid of the `console.log`. 🗡

routes/personal/personal.spec.js
```diff
@@ -1,9 +1,43 @@
...
+
+  })
```
pcraig3: It would be good to test:

- expect status (422) for `.post()` with no marital status
- expect status (422) for `.post()` with fake marital status
- expect status (302) for `.post()` with good marital status

utils.js
```diff
@@ -19,6 +19,20 @@ const errorArray2ErrorObject = (errors = []) => {
...
+//POST functions that handle setting the login data in the session and handle redirecting to the next page or sending an error to the client.
+//Note that this is not the only error validation, see routes defined above.
+const validateRedirect = (req, res, next) => {
+  let redirect = req.body.redirect || null
+  
+  if (!redirect) {
+    throw new Error(`[POST ${req.path}] 'redirect' parameter missing`)
+  }
+  return next()
+}
```
pcraig3: Nice! very clean. 😗👌

views/personal/maritalStatus-edit.pug
```diff
@@ -0,0 +1,47 @@
...
+    div
+      details.
+        <summary><span>Help with editing marital status</span></summary>
+        <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.</p>
+
+    input#redirect(name='redirect', type='hidden', value='personal/maritalStatus')
+
+    .buttons
+      .buttons--left
+        button(type='submit') Change Marital Status
```
pcraig3: Not a "right now" thing, but I think we should make the left button bigger when the text is bigger 🤔

katedee: agreed! it's the flex setting on those buttons right now, but I'll take a look at it!

formSchemas.js
```diff
@@ -38,7 +38,17 @@ const sinSchema = {
   },
 }
 
+const maritalStatusSchema = {
+  maritalStatus: {
+    isLength: {
+      errorMessage: 'errors.maritalStatus.maritalStatus',
+      options: { min: 1 }
+    }
```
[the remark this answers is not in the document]
pcraig3: Here's an example: https://github.com/cds-snc/cra-alpha/blob/master/src/questions/maritalStatus.js#L16

public/scss/_forms.scss
sboisvert: What does this do?

routes/login/login.controller.js
```diff
@@ -70,8 +57,8 @@ const postSIN = (req, res) => {
   }
 
   //Success, we can redirect to the next page
-  if (sin && redirect) {
-    return res.redirect(redirect)
+  if (sin) {
+    return res.redirect(req.body.redirect)
```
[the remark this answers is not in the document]
sboisvert: lol, good call

routes/personal/personal.spec.js
```diff
@@ -1,9 +1,60 @@
...
+      expect($('input[name=maritalStatus]:checked').val().toLowerCase).toEqual(user.personal.maritalStatus.toLowerCase)
+    })
+
+    test('it returns a 422 with no marital status', async () => {
+      const response = await request(app).post('/personal/maritalStatus/edit').send({ redirect: '/personal/maritalStatus'})
+      expect(response.statusCode).toBe(422)
+    })
+
+    test('it returns a 422 with fake marital status', async () => {
+      const response = await request(app).post('/personal/maritalStatus/edit').send({ maritalStatus: 'cat lady', redirect: '/personal/maritalStatus'})
```
pcraig3: 😸

PR #20, opened 2019-07-10
Short term fix for errors

sboisvert:
Right now errors fail out, meaning we get errors about the error handling code instead of the actual error. I'm not 100% sure this fixes everything, but it makes the error slightly more understandable. It could potentially be improved cc @ayoajila

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,5 +1,6 @@
 // import environment variables.
 require('dotenv').config()
+const globalError = require('http-errors');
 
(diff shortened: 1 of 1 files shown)


pcraig3:
@sboisvert, I took your branch and ran the linter: `npm run lint`. 

If you install the `eslint` plugin for VSCode, there's a setting where it autolints on save.

PR #21, opened 2019-07-10
feature/Deductions

sboisvert:
Stub file for claims controller and spec file for tests. Also includes a stub rrsp file in views.


routes/claims/claims.controller.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+    return redirect
+}
+
+const postRRSP = (req, res) => {
+    const redirect = validateRedirect(req, res)
+
+    const errors = validationResult(req)
+
+    //If sin is not set, set it to null
```
pcraig3: comment still refers to 'sin'

routes/claims/claims.controller.js
```diff
@@ -0,0 +1,50 @@
...
+    let rrsp = req.body.rrsp || null
+    req.session.credits.rrspClaim = rrsp
```
pcraig3: Is this going to be `deductions` now, instead of credits?

routes/claims/claims.controller.js
```diff
@@ -0,0 +1,50 @@
...
+        })
+    }
+
+    //Success, we can redirect to the next page
+    if (rrsp && redirect) {
+        return res.redirect(redirect)
+    }
+
+    //No errors, but didn't process. Unsure if this code branch is reachable but it's a good safety.
+    res.status(422).render('deductions/rrsp', { data: req.session || {} })
```
pcraig3: Yeah, I'm going to refactor these later. But for now this is a good catch.

routes/claims/claims.controller.js
```diff
@@ -0,0 +1,50 @@
+const { validationResult, checkSchema } = require('express-validator')
```
pcraig3: The `claims.controller` is handling the `/deductions` routes, and the `deductions.controller` is doing the opposite. Can we just have one controller (deductions) that handles them all?

routes/deductions/deductions.spec.js.js
```diff
@@ -0,0 +1,3 @@
+const request = require('supertest')
+const cheerio = require('cheerio')
+const app = require('../../app.js')
```
pcraig3: One test for a 200 response and one for the redirect would be good.

routes/claims/claims.controller.js
sboisvert: ya, not sure how I missed that one

routes/claims/claims.controller.js
sboisvert: fixed

routes/claims/claims.controller.js
sboisvert: yes

routes/deductions/deductions.spec.js.js
sboisvert: done!

routes/deductions/deductions.controller.js
```diff
@@ -0,0 +1,50 @@
...
+}
+
+const postRRSP = (req, res) => {
+  const redirect = validateRedirect(req, res)
+
+  const errors = validationResult(req)
+
+  //If rrsp is not set, set it to null
+  let rrsp = req.body.rrsp || null
+  req.session.deductions.rrspClaim = rrsp
```
pcraig3: We shouldn't be updating the session until we're past the `(!errors.isEmpty())` check. But otherwise, yeah, this is the way to update user session stuff.

pcraig3: One thing to note is that I've stubbed out all the user session deductions/claims.

So, two things.

1. [Update `"claims":` to `"deductions":`](https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/api/user.json#L51)

2. Since lots of them are opt-in, I've added the `{whatever}Claim` key meant to be true/false (or null as the default), but then we can add more keys as we need for the data the user is actually submitting. In this case, [if `rrspClaim` is true](https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/api/user.json#L53), then we should expect [an `rrspAmount`](https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/api/user.json#L54) as well.

routes/deductions/deductions.controller.js
```diff
@@ -0,0 +1,50 @@
...
+
+//POST functions that handle setting the login data in the session and handle redirecting to the next page or sending an error to the client.
+//Note that this is not the only error validation, see routes defined above.
+const validateRedirect = req => {
+  let redirect = req.body.redirect || null
+  if (!redirect) {
+    throw new Error(`[POST ${req.path}] 'redirect' parameter missing`)
+  }
+  return redirect
+}
```
pcraig3: Now that @katedee's redirect-checking middleware is in, you can include it [in the chain of callbacks that are part of the post request](https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/routes/login/login.controller.js#L15), and delete this `validateRedirect` function.

routes/deductions/deductions.controller.js
sboisvert: done

routes/deductions/deductions.controller.js
sboisvert: Good call

routes/deductions/deductions.controller.js
```diff
@@ -0,0 +1,40 @@
+const { validationResult } = require('express-validator')
+const { errorArray2ErrorObject, validateRedirect } = require('./../../utils.js')
+//const { loginSchema, sinSchema } = require('./../../formSchemas.js')
+//const API = require('../../api')
```
katedee: if we're not using these, let's just removed the lines for now

routes/deductions/deductions.spec.js
```diff
@@ -0,0 +1,26 @@
+const request = require('supertest')
+//const cheerio = require('cheerio')
```
katedee: same here, remove if not using

PR #22, opened 2019-07-11
[WIP] Feature/Sin looks at user's actual social insurance number

pcraig3:
Part of the "login" flow means checking the SIN against the actual SIN of the user trying to sign in, not just validating it has the correct format.

Added logic such that:

- if there _is_ a user session, check that the SIN is the correct one
- if there _isn’t_ a user session, just validate the format

I think probably we want to KO the second case at some point soon, but for now we still let people see the SIN page without a real access code, and it will still send them to the success page.

### I also updated updated the SIN page with user's name and SIN in the hint text

Since we're going to be (eventually) creating different user journeys through the application, asking someone demoing the service to enter a correct

- access code
- SIN
- date of birth

is a pretty big lift, even though that's how the app works.

To make it easier, I updated the hint text so that it shows the correct value for the current user.

This is just a "dev" feature, so we will revert to static hint text once we have a live service.

## gif

sin

### DIFF
--- a/routes/login/login.controller.js
+++ b/routes/login/login.controller.js
@@ -39,7 +39,6 @@ const postLoginCode = (req, res) => {
 }
 
 const postSIN = (req, res) => {
-
   //If sin is not set, set it to null
   let sin = req.body.sin || null
   req.session.sin = sin
@@ -47,9 +46,6 @@ const postSIN = (req, res) => {
   const errors = validationResult(req)
 
   if (!errors.isEmpty()) {
-    // clear session
-    req.session = null
-
     return res.status(422).render('login/sin', {
       data: { sin: req.body.sin } || {},
       errors: errorArray2ErrorObject(errors),
--- a/routes/login/login.controller.js
+++ b/routes/login/login.controller.js
@@ -43,7 +43,7 @@ const postSIN = (req, res) => {
 
   if (!errors.isEmpty()) {
     return res.status(422).render('login/sin', {
-      data: { sin: req.body.sin } || {},
+      data: { ...req.session, ...{ sin: req.body.sin } } || {},
       errors: errorArray2ErrorObject(errors),
     })
   }
--- a/routes/login/login.controller.js
+++ b/routes/login/login.controller.js
@@ -1,4 +1,4 @@
-const { body, validationResult, checkSchema } = require('express-validator')
+const { validationResult, checkSchema } = require('express-validator')
 const { errorArray2ErrorObject, validateRedirect } = require('./../../utils.js')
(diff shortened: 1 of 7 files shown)


pcraig3:
@sboisvert @katedee if we're okay with this way of doing things, I was going to add few session tests ([like this](https://github.com/cds-snc/cra-alpha/blob/master/src/__tests__/session.test.js#L85)) and call it a day.

We could also add functional tests to it (ie, using cypress), but it still feels like we don't have enough of a journey built out to justify it.

routes/login/login.controller.js
```diff
@@ -39,19 +39,11 @@ const postLoginCode = (req, res) => {
...
-
   const errors = validationResult(req)
 
   if (!errors.isEmpty()) {
-    // clear session
-    req.session = null
-
     return res.status(422).render('login/sin', {
-      data: { sin: req.body.sin } || {},
+      data: { ...req.session, ...{ sin: req.body.sin } } || {},
```
sboisvert: What are the benefits of passing it as a spread instead of as an array?

views/login/sin.pug
```diff
@@ -7,14 +7,21 @@ block content
...
   p Now that we have your personal code, we need some additional information in order to verify that it’s you.
 
   p Please enter your Social Insurance Number (SIN) in the field below.
 
   form.cra-form(method='post')
     div
       label(for='sin') Social Insurance Number
-      span.cra-form-message For Example: 123 456 789
+      if data.personal
+        span.cra-form-message For Example: #{SINFilter(data.personal.sin)}
```
katedee: I don't know why my brain didn't realize I could pull functions in directly to pug templates, but this is nice 👌

sboisvert: I also didn't know this!

routes/login/login.controller.js
pcraig3: Spreads let you merge dictionaries — arrays don't let you do that.

So, for example:

```js
var dict1 = {
  a: 'A',
  b: 'B'
}

var dict2 = { ...dict1, ...{ c: 'C' } }

/* 

 dict2 looks like this 

{
  a: 'A',
  b: 'B',
  c: 'C'
}
*/
```

In this case, we need to pass in 

- the value that the user gave us (so we can put it back in the input)
- the good value (ie, for the hint text)

So we merge both dictionaries together.

views/login/sin.pug
pcraig3: Yeah, Pug's way of doing "filters" only works for static text, so this is probably what we want most of the time.

PR #23, opened 2019-07-11
[WIP] - Feature/radio includes

katedee:
This is a base idea around how to handle radio input forms in our express/pug app.

Plenty of room for improvement, I'm sure. I've left some comments in areas about things that may feel clunky/reasoning behind, but am open to suggestions (even if the suggestion is to scrap it)

having issues with one of tests right now, will sort that out later if we feel this is worth going ahead with

The idea is:
- keep the radio form values (which is essentially text for our labels) in a file that can be imported multiple places, since we'll need these values in multiple places (validation, text, etc), and it'd be nice to have one source of truth vs. editing copy in multiple files
- pass these values along to our route, access them in the pug file
- pug file has a var which is an object with the common vals in the radio forms (names for buttons, dataKey which is used to check if the button should be pre-checked on page load), and then a list of values we're using for the radio buttons

### DIFF
--- a/formRadios.js
+++ b/formRadios.js
@@ -1,0 +1,10 @@
+const maritalStatusForm = {
+    common: { 
+        name: 'maritalStatus', 
+        dataKey: (data && data.personal && data.personal.maritalStatus ? data.personal.maritalStatus : null) 
+    }, 
+    values: ['Single','Married','Separated', 'Divorced', 'Widowed']
+}
+module.exports = {
+  maritalStatus
+}
--- a/formRadios.js
+++ b/formRadios.js
@@ -1,7 +1,28 @@
 // haven't used this elsewhere yet, but the idea of keeping it here is so we can pull these values in wherever we're using them, and have one source of truth for these values (instead of updating copy in multiple files)
-const maritalStatusValues = {
-    en: ['Single','Married','Separated', 'Divorced', 'Widowed']
+const maritalStatusOptions = {
+    en: [
+      {
+        value: 'Single',
+        label: 'Single',
+      },
+      {
+        value: 'Married',
+        label: 'Married',
+      },
+      {
+        value: 'Separated',
+        label: 'Separated',
+      }, 
+      {
+        value: 'Divorced',
+        label: 'Divorced',
+      }, 
+      {
+        value: 'Widowed',
+        label: 'Widowed',
+      }
+    ]
 }
 module.exports = {
-  maritalStatusValues
+  maritalStatusOptions
 }
--- a/formRadios.js
+++ b/formRadios.js
@@ -1,3 +1,4 @@
+// haven't used this elsewhere yet, but the idea of keeping it here is so we can pull these values in wherever we're using them, and have one source of truth for these values (instead of updating copy in multiple files)
 const maritalStatusValues = {
     en: ['Single','Married','Separated', 'Divorced', 'Widowed']
 }
--- a/formRadios.js
+++ b/formRadios.js
@@ -1,10 +1,6 @@
-const maritalStatusForm = {
-    common: { 
-        name: 'maritalStatus', 
-        dataKey: (data && data.personal && data.personal.maritalStatus ? data.personal.maritalStatus : null) 
-    }, 
-    values: ['Single','Married','Separated', 'Divorced', 'Widowed']
+const maritalStatusValues = {
+    en: ['Single','Married','Separated', 'Divorced', 'Widowed']
 }
 module.exports = {
-  maritalStatus
+  maritalStatusValues
 }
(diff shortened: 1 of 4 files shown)


views/personal/maritalStatus-edit.pug
```diff
@@ -10,27 +10,18 @@ block content
   div
     p Please choose the correct status from the options listed below.
 
+  //- this is messy (data && data.personal && data.personal.maritalStatus) but pug will error if a property doesn't exist. We could handle this on the controller level, and pass this to the route? Will help keep it a bit cleaner here
```
pcraig3: Yeah, I hate that

views/personal/maritalStatus-edit.pug
```diff
@@ -10,27 +10,18 @@ block content
...
+  - var radioFormOther = {common: { name: 'maritalStatus', dataKey: (data && data.personal && data.personal.maritalStatus ? data.personal.maritalStatus : null) }, options: maritalStatusOptions.en }
```
pcraig3: Is `dataKey` the same as the "value" html attribute? (ie, could we call it `value`)

formRadios.js
```diff
@@ -0,0 +1,28 @@
+// haven't used this elsewhere yet, but the idea of keeping it here is so we can pull these values in wherever we're using them, and have one source of truth for these values (instead of updating copy in multiple files)
+const maritalStatusOptions = {
+    en: [
+      {
+        value: 'Single',
+        label: 'Single',
+      },
+      {
+        value: 'Married',
```
pcraig3: Currently we have this as "Married or common-law"

formRadios.js
```diff
@@ -0,0 +1,28 @@
...
+      }, 
+      {
+        value: 'Widowed',
+        label: 'Widowed',
+      }
+    ]
+}
+module.exports = {
+  maritalStatusOptions
+}
```
pcraig3: HMMMMM.

Can we put these in the schemas file somewhere? 
That way everything to do with questions will live together.

Also, the `en`, `fr` thing I am a bit suspicious of (all of our other content will work with the translate lib right now), but happy to let this through if you think it's easier than the alternative (ie, like what we've done with error messages). 

Finally, we wouldn't want to change the value in French, only the key.

pcraig3: If you wanted to keep it in its own file, you could do something like I did with the /utils file recently. (ie, move utils.js to /utils/index.js)

https://github.com/cds-snc/cra-claim-tax-benefits/pull/25/files#diff-8d6b1f9270334e293ad1f2ddec84e050

views/personal/maritalStatus-edit.pug
katedee: so datakey is checking the data as we enter the page it's basically used to run the check:
`checked=(radioForm.common.dataKey &&radioForm.common.dataKey.toLowerCase()==option.value.toLowerCase()))` to see if the item should be checked when we load the page (ex. someone updating their marital status where previously it was 'Single' and so when they get to the page 'Single' will be checked before they change it).

We could get rid of pre-checking altogether, or handle it differently! I'm open to alternatives as making sure that first `data` exists, and then `data.personal` exists, and then `data.personal.maritalStatus` exists. We could offload this series of checks into a function that's in utils?

formRadios.js
katedee: will update!

views/personal/maritalStatus-edit.pug
sboisvert: >Will help keep it a bit cleaner here

Trying to keep the view as clean as possible seems reasonable.

views/personal/maritalStatus-edit.pug
```diff
@@ -10,27 +10,18 @@ block content
...
-                    label.radios__label(for="maritalStatusSeparated") Separated
-                .radios__item
-                    input.radios__input(id="maritalStatusDivorced" name="maritalStatus" type="radio" value="Divorced" checked=(data && data.personal.maritalStatus.toLowerCase()=='divorced') aria-describedby=(errors ? 'maritalStatus_error' : false))
-                    label.radios__label(for="maritalStatusDivorced") Divorced
-                .radios__item
-                    input.radios__input(id="maritalStatusWidowed" name="maritalStatus" type="radio" value="Widowed" checked=(data && data.personal.maritalStatus.toLowerCase()=='widowed') aria-describedby=(errors ? 'maritalStatus_error' : false))
-                    label.radios__label(for="maritalStatusWidowed") Widowed
+        
+            include ../_includes/radioButtons.pug
+            +radio-buttons(radioFormOther)
```
sboisvert: What does `radioFormOther` reference?

PR #26, opened 2019-07-12
Fixup/Improvements to validation messages

pcraig3:
Bit of an overhaul here. Let's go through them one by one.

## 1. Add "Error: " to page titles when there is an error 
This is in line with the GDS error advice.

> add ‘Error: ’ to the beginning of the <title> so screen readers read it out as soon as possible
> -- https://design-system.service.gov.uk/components/error-summary/#how-it-works

## 2. Add validation message mixin

Included a validation message mixin in the base file as an `include` statement, which means that all of our other templates have access to it.

Prefixed all validation messages with a visually hidden "Error: " string, also in line with the GDS guidance on validation messages.

> To help screen reader users, the error message component includes a hidden ‘Error:’ before the error message. These users will hear, for example, “Error: The date your passport was issued must be in the past”.
> -- https://design-system.service.gov.uk/components/error-message/

Also added a `.visually-hidden` class to the repo so that we can make content available to screenreaders but not visual users. [I have an 18F link](https://accessibility.18f.gov/hidden-content/) that sums this up, and it's basically the same CSS.

## 3. Move validation messages above inputs 

We mulled it over with Julianna and here's where we came down.

> P: My assumption is move the error messages between the label and the input.
> J: Sometimes its more complicated but generally speaking its a good method.

The GDS guidance more or less backs this up, although it's not 1000% explicit.

> put the message in red after the question text and hint text
> -- https://design-system.service.gov.uk/components/error-message/

So it doesn't say "before the input", but that's how it is in all of their examples.

Since our error messages show up after the page reloads, the reading order is assumed to be

```
label
hint text
validation message
input
```

So it works for both visual users and screen reader users.

## 4. Point aria-described by at the validation message

The validation message is a bit closer to the action, and now it has a secret "Error: " prefix as well.

It's probably a bit more reliable this way.

## 5. Remove "{id}_error" `id` from summary list items

Now that the aria-describedby is pointing at the validation messages (see: 4), we don't need these ids in the error summary list items.

## 6. Add chunky red left-border to inputs with validation errors

Talked it over with the designers and this was the outcome.

Basically (as with everything else, basically), the GOV.UK errors are really good, so let's do what they do.

> use a red border to visually connect the message and the question it belongs to
> -- https://design-system.service.gov.uk/components/error-message/#how-it-works

# Screenshots 

| before | after |
|--------|-------|
| Screen Shot 2019-07-11 at 10 40 23 PM  |  Screen Shot 2019-07-11 at 10 39 36 PM   |
|  Screen Shot 2019-07-11 at 10 41 11 PM   |  Screen Shot 2019-07-11 at 10 39 51 PM  |


routes/login/login.spec.js
```diff
@@ -41,19 +41,20 @@ describe('Test /login responses', () => {
         .post('/login/code')
         .send({ redirect: '/' })
       const $ = cheerio.load(response.text)
+      expect($('title').text()).toMatch(/^Error:/)
       expect($('.error-list__header').text()).toEqual('Please correct the errors on the page')
       expect($('.error-list__list').children()).toHaveLength(1)
-      expect($('.validation-message').text()).toEqual('Access code must be 8 characters')
-      expect($('#code').attr('aria-describedby')).toEqual('code_error')
+      expect($('.validation-message').text()).toEqual('Error: Access code must be 8 characters')
+      expect($('#code').attr('aria-describedby')).toEqual('code--error')
```
katedee: what about having this class just be `code-error`? `--` is BEM thing which assumes this is a variant of code

pcraig3: Sure, that works.

PR #31, opened 2019-07-12
Add hasData function to use on templates

katedee:
Add hasData function to use on templates
so we don't have to clutter with those long checks like `data && data.personal && data.address`, etc

If we think this is good, I'll add just a few more tests to it :)

Let me know what you think!


pcraig3:
omg I love it

PR #33, opened 2019-07-14
Feature/dob

katedee:
I underestimated the validation this needed 😅

### TL;DR:
- created Date of Birth Page
- updated user.json to have dob formatted as we are taking it `2019/02/02` vs `2019-02-02`
- added some code in formSchemas that allow for multiple custom validations
- validations...so...many...validations
- added a bunch of tests
- sin page redirects to dob

### validations included
- make sure date has been entered in the correct format (years is 4 characters, month is 2, day is 2)
- with slashes, make sure the input is not longer or less than 10 characters
- no characters beyond numbers and slashes
- make sure date is no greater than a year ago (must be at least 1yr old...and yes we can change this to be older later)
- make sure the person is not older than 200y/o. Honestly, if you live to 201, I think you should be allowed to stop filing taxes. Again we can change this to what we feel is more reasonable, or get rid of it altogether
- make sure it's a valid month
- make sure it's a valid day within that month
- date of birth matches to corresponding access code

And tests, to test all of that

image


formSchemas.js
```diff
@@ -50,6 +80,106 @@ const sinSchema = {
...
+      validBirthDateChars,
+      validMonth,
+      validDay,
+      validYear,
+      isMatchingDoB,
+    ]),
+    isLength: {
+      errorMessage: 'errors.login.dateOfBirth',
+      //length with slashes
+      options: { min: 10, max: 10 },
```
pcraig3: This means that people have to do `01` for days less than 10, right?

Like `1990/10/02`?

locales/fr.json
```diff
@@ -9,6 +9,13 @@
   "errors.login.lengthSIN": "Your SIN should have 9 numbers",
   "errors.maritalStatus.maritalStatus": "You must select one of the options",
   "errors.login.sin": "SIN does not match access code",
+  "errors.login.dateOfBirth": "Please enter a valid birth date",
+  "errors.login.dateOfBirth.match": "Date of birth does not match access code",
+	"errors.login.dateOfBirth.format": "Make sure the date is formatted as yyyy/mm/dd",
+	"errors.login.dateOfBirth.characters": "The date may only contain numbers and slashes. Please format as yyyy/mm/dd",
+	"errors.login.dateOfBirth.validMonth": "Please enter a valid month between 01 and 12",
+	"errors.login.dateOfBirth.validDay": "Please enter a valid day",
+	"errors.login.dateOfBirth.validYear": "Please enter a valid year",
```
pcraig3: The formatting is a bit weird, but these are auto generated anyway, so w/e.

routes/login/login.spec.js
```diff
@@ -193,9 +193,167 @@ describe('Test /login responses', () => {
...
+
+    const currentDate = new Date()
+    const year = currentDate.getFullYear()
+    const month = currentDate.getMonth()
+    const day = currentDate.getDate()
+
+    const dateToString = (dateInput) => {
+      const add0 = (s) => { return (s < 10) ? '0' + s : s }
+      const d= new Date(dateInput)
+      return [d.getFullYear(),add0(d.getMonth()+1), add0(d.getDate())].join('/')
```
pcraig3: `+1`

Dates in javascript are terrible 😧

formSchemas.js
```diff
@@ -50,6 +80,106 @@ const sinSchema = {
...
+    },
+    
+  }
+}
+
```
pcraig3: On the one hand, good work 👏 looks great!

On the other, if this starts happening again, ping me. It's a lot of work for something that we don't have a super strong opinion about.

I left a comment in the card (I think, not visibly enough) about how we should just default to ISO dates, and [there's an ISO validator in the validation lib we're using](https://github.com/validatorjs/validator.js#validators) (it's called `isISO8601(str)`), so we probably could have gotten away with that. 

But, in any case, looks good and well-tested, so A+! 👍

formSchemas.js
```diff
@@ -1,5 +1,35 @@
...
+						const errorMessage = validator.errorMessage || `${value} is invalid`;
+						errors.push(errorMessage);
+					}
+					return result;
+				});
+				return results.every(value => value === true);
+			}
+		}
+	};
+};
```
pcraig3: I think your editor likes 4 spaces and mine likes 2. Can we standardize on 2? 

I'm going to push another commit on here.

routes/login/login.controller.js
```diff
@@ -13,6 +13,10 @@ module.exports = function(app) {
   //SIN
   app.get('/login/sin', (req, res) => res.render('login/sin', { data: req.session || {} }))
   app.post('/login/sin', validateRedirect, checkSchema(sinSchema), postSIN)
+
+  //Date of Birth
+  app.get('/login/dateOfBirth', (req, res) => res.render('login/dob', { data: req.session || {} }))
+  app.post('/login/dateOfBirth', validateRedirect, checkSchema(birthSchema), postDoB)
```
pcraig3: I think we should either

- change the URL to `/dob` to match the template
- change the template to `dateOfBirth.pug` to match the URL

formSchemas.js
pcraig3: (yes, just saw the test)

formSchemas.js
pcraig3: I'm using VSCode. Are you using it as well or a different one?

formSchemas.js
katedee: we could change this, and quickly sanitize the value to see if it needs a `0` beforehand, if the user hasn't entered it

formSchemas.js
katedee: ahahaha oops

formSchemas.js
katedee: using it the same, I keep trying to keep 2 spaces, but it protests! will clean it up though

PR #34, opened 2019-07-15
Feature/Use user data + make sure it's always there

pcraig3:
This is a pretty important one.

**TL;DR**

- Created a new middleware to inject user data into the session for all pages after "/login/code"
- Linked (nearly) all the finished pages together that we've built
- Created a new "body" key in the couple places where we pass user data back into the template

More below, but the point of adding this middleware is that we don't have to fake out data for our pages when we're building them.

##  Add checkPublic middleware to the app

Since we need user data to be in the app for most pages, and since we're rapidly devving, testing, and demoing pages, we should guarantee that we have user data in the session for all the non-public paths.

Once this is handling real user data, we can change this middleware to redirect everyone to the start page if they try to manually go to a URL but haven't signed in. But for right now, it's more useful for us to make sure the session is there rather than locking people out of the app.

A knock-on effect of this is that we can remove a bunch of "if hasData" checks from our templates now and going forward. 

##  Add user-submitted address parameters to the page under the "body" key

Basically the problem is:

- we want to keep the original user address around
- but we want to be able to preserve the value the user typed in instead of just losing it

So by passing it in as "body", we can do both.

The "if" check for setting variables at the beginning of the file
will do:

- use `body[key]` if exists, and isn't null or empty string
- use `data[key]` if exists, and isn't null or empty string
- use ""

It looks really gnarly but I think it's actually a Big Improvement ™.

Also sent the `sin` and `dateOfBirth` back as a "body" key.

### DIFF
--- a/api/user.json
+++ b/api/user.json
@@ -6,7 +6,7 @@
     "firstName": "Gabrielle",
     "lastName": "Roy",
     "sin": "111222333",
-    "dob": "1909/03/22",
+    "dateOfBirth": "1909/03/22",
     "address": {
       "aptNumber": "",
       "streetNumber": "375",
@@ -26,7 +26,7 @@
     "firstName": "Marcel",
     "lastName": "Carbotte",
     "sin": "222333444",
-    "dob": "1914/02/09",
+    "dateOfBirth": "1914/02/09",
     "disability": false,
     "netIncome": "$1,000.00"
   },
@@ -34,7 +34,7 @@
     {
       "firstName": "Bernadette",
       "lastName": "Roy",
-      "dob": "1948/09/25",
+      "dateOfBirth": "1948/09/25",
       "disability": false,
       "netIncome": "0",
       "relationship": "child"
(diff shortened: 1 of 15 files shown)


api/user.json
```diff
@@ -6,7 +6,7 @@
     "firstName": "Gabrielle",
     "lastName": "Roy",
     "sin": "111222333",
-    "dob": "1909/03/22",
+    "dateOfBirth": "1909/03/22",
     "address": {
```
katedee: 👌

views/personal/maritalStatus-edit.pug
```diff
@@ -18,19 +19,19 @@ block content
                 if errors
                   +validationMessage(errors.maritalStatus.msg, 'maritalStatus')
                 .radios__item
-                    input.radios__input(id="maritalStatusSingle" name="maritalStatus" type="radio" value="Single" checked=(data && data.personal.maritalStatus.toLowerCase()=='single') aria-describedby=(errors ? 'maritalStatus-error' : false))
+                    input.radios__input(id="maritalStatusSingle" name="maritalStatus" type="radio" value="Single" checked=(maritalStatus.toLowerCase()=='single') aria-describedby=(errors ? 'maritalStatus-error' : false))
```
katedee: much cleaner, ty ✨

PR #35, opened 2019-07-15
Feature/RRSP page

pcraig3:
Took over @sboisvert's existing RRSP page and finagled it into better shape. 

This is the simplest of a common pattern that's in the current wireframes, so we need to get this out sooner rather than later so that we can pump out the subsequent pages more quickly.

We were missing the "yes/no" page that we want to ask initially. If they say no, we don't need to ask them for the amount that they spent on RRSPs (duh). 

- Added Yes/No page + tests
- Added an input with a dollar sign — it comes up a bunch in our later designs as well
  - Added a "currency" schema that allows things that look `$10`, `10`, or `10.00`
- Also changed our form labels so that they use "Lato" Sans like our page titles

## Screenshots

rrsp


public/scss/components/_radios.scss
```diff
@@ -0,0 +1,71 @@
+.radios {
+  $radioClass: &;
+
+  &__item {
+    font-size: 19px;
+    font-size: 1.1875rem;
```
katedee: I was thinking of a separate file, nice 😎

PR #36, opened 2019-07-15
Province of residence page 

nmakuch:
### Creating the province of residence page
Residence

### Breakdown
- Created this page based off of the Marital status page
- Added `personal/residence` route
- Created `residenceSchema`
- Created the logic to redirect based on response
- Right now `Yes` will redirect you to the address page and
- `No` will redirect you to the start page (this will be switched to an off boarding page once it has been developed)
- Hoping I got the server logic correct (@pcraig3 don't know if needed to add a status code or something?)
- Basic tests written for submitting `Yes (302)`, `No (302)`, `No answer (422)`

### Trello card
[Create page for Province Lived in past year](https://trello.com/c/gCgF9nu2)

### DIFF
--- a/formSchemas.js
+++ b/formSchemas.js
@@ -268,7 +268,7 @@ const residenceSchema = {
   residence: {
     isIn: {
       errorMessage: 'errors.residence',
-      options: [[ 'yes' , 'no' ]],
+      options: [[ 'Yes' , 'No' ]],
     },
   },
 }
--- a/formSchemas.js
+++ b/formSchemas.js
@@ -268,7 +268,7 @@ const residenceSchema = {
   residence: {
     isIn: {
       errorMessage: 'errors.residence',
-      options: [[ true , false ]],
+      options: [[ 'yes' , 'no' ]],
     },
   },
 }
--- a/formSchemas.js
+++ b/formSchemas.js
@@ -264,9 +264,19 @@ const rrspAmountSchema = {
   },
 }
 
+const residenceSchema = {
+  residence: {
+    isIn: {
+      errorMessage: 'errors.residence',
+      options: [['Alberta', 'British Columbia', 'Manitoba', 'Newfoundland and Labrador', 'Northwest Territories', 'Nova Scotia', 'Nunavut', 'Ontario', 'Prince Edward Island', 'Quebec', 'Saskatchewan', 'NonResident', '0']],
+    },
+  },
+}
+
 module.exports = {
   loginSchema,
   maritalStatusSchema,
+  residenceSchema,
   sinSchema,
   addressSchema,
   birthSchema,
--- a/formSchemas.js
+++ b/formSchemas.js
@@ -268,7 +268,7 @@ const residenceSchema = {
   residence: {
     isIn: {
       errorMessage: 'errors.residence',
-      options: [['Alberta', 'British Columbia', 'Manitoba', 'Newfoundland and Labrador', 'Northwest Territories', 'Nova Scotia', 'Nunavut', 'Ontario', 'Prince Edward Island', 'Quebec', 'Saskatchewan', 'NonResident', '0']],
+      options: [[ true , false ]],
     },
   },
 }
--- a/routes/personal/personal.controller.js
+++ b/routes/personal/personal.controller.js
@@ -96,8 +96,7 @@ const postResidence = (req, res) => {
     residence: residence,
   }
 
-  console.log('req.body', req.body)
-  console.log('errors', errors)
+  console.log(req.body.residence)
 
   if (!errors.isEmpty()) {
     return res.status(422).render('personal/residence', {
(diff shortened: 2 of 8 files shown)


routes/personal/personal.controller.js
```diff
@@ -9,6 +9,14 @@ module.exports = function(app) {
   app.get('/personal/address/edit', getAddressEdit)
   app.post('/personal/address/edit', validateRedirect, checkSchema(addressSchema), postAddress)
 
+  app.get('/personal/residence', (req, res) => res.render('personal/residence'))
+  app.post(
+    '/personal/residence',
+    checkSchema(residenceSchema),
+    validateRedirect,
```
pcraig3: Let's move this in front of the schema check. If the `redirect` isn't there, we should fail that first before checking the schema at all.

views/personal/residence.pug
```diff
@@ -0,0 +1,29 @@
...
+
+  form.cra-form(method='post')
+    +yesNoRadios('residence', residence, 'Ontario resident on Dec.31, 2018?', errors)
+
+    div
+      details.
+        <summary><span>Help with Ontario residence</span></summary>
+        <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.</p>
+
+    input#redirect(name='redirect', type='hidden', value='/personal/residence')
```
katedee: this should redirect to the page we want to go to next (assuming a successful answer). I think I saw in your controller earlier that you want to direct to the address page?

routes/personal/personal.controller.js
```diff
@@ -79,3 +87,25 @@ const postMaritalStatus = (req, res) => {
...
+      data: { residence: req.body.residence } || {},
+      errors: errorArray2ErrorObject(errors),
+    })
+  }
+
+  if (errors.isEmpty() && req.body.residence === 'Yes') {
+    return res.redirect('address')
+  } else {
+    return res.redirect('/start')
+  }
```
katedee: This is good, but just to be consistent with the style we've set, what do you think of:
```
if (req.body.residence !== 'Yes') {
 return res.redirect('/start')
}

return res.direct(req.body.redirect)
```

`req.bod.redirect` being the value we put in the hidden input on the page itself (mentioned in another comment)

routes/personal/personal.controller.js
```diff
@@ -79,3 +87,25 @@ const postMaritalStatus = (req, res) => {
...
   return res.redirect(req.body.redirect)
 }
+
+const postResidence = (req, res) => {
+  const errors = validationResult(req)
+
+  let residence = req.body.residence || null
+  req.session.personal = {
+    residence: residence,
+  }
```
katedee: we don't want to assign until we know things are successful, so let's pop this down after the `if(!errors.isEmpty())` check.
Also we can do away with the `let` up there since we're not manipulating the data into any shape. 
We also don't want to override the whole object, so I think we can change lines 94-97 to one line which is just `req.session.personal.residence = req.body.residence`
We updated how we're doing this recently, so if you look at the latest version of master you can see maritalStatus has been updated like this as well

formSchemas.js
```diff
@@ -264,9 +264,19 @@ const rrspAmountSchema = {
   },
 }
 
+const residenceSchema = {
+  residence: {
+    isIn: {
+      errorMessage: 'errors.residence',
+      options: [[ 'Yes' , 'No' ]],
+    },
```
katedee: 👌

routes/personal/personal.controller.js
nmakuch: going to remove this actually. Had a chat with @pcraig3 and we aren't actually going to be saving this value anywhere, just using it to redirect to corresponding pages.

pcraig3: Yeah, axe it.

routes/personal/personal.spec.js
```diff
@@ -65,6 +66,29 @@ describe('Test /personal responses', () => {
...
+    })
+
+    test('it returns a 302 when selecting NO', async () => {
+      const response = await request(app)
+        .post('/personal/residence')
+        .send({ redirect: '/', residence: 'No' })
+      expect(response.statusCode).toBe(302)
+    })
+
+    test('it returns a 302 when selecting YES', async () => {
```
pcraig3: You _could_ write one more test for some bad value (ie, not 'Yes', or 'No'), but I already wrote a bunch somewhere else and it would just be boilerplate.

So, on second thought, this is fine.

views/personal/residence.pug
```diff
@@ -0,0 +1,29 @@
+extends ../base
+include ../_includes/yesNoRadios
+
+block variables
+  -var title = 'Confirm if you are an Ontario resident'
```
pcraig3: You can no longer be an Ontario resident and still use this (as long as you lived here last year). But this is more clear for the purposes of our demo so it's cool.

PR #37, opened 2019-07-15
Task/Add Yes+No radios Mixin

pcraig3:
I'm thinking we should standardize these because they show up in a few places.

For example, there's one in #36.


pcraig3:
Yeah, let's do it!

PR #38, opened 2019-07-15
Feature/Error checking middleware to reduce boilerplate code

pcraig3:
@katedee, what do you think? 

## Reduce boilerplate by putting error checking in middleware

Since most of our error checks are pretty similar, I built out a middleware that runs through the error check. If it finds an error, it 

- returns a `422` status
- passes in the session as `data`
- passes in any request parameters as `body`
- passes in the errors as `errors`

## Add extra check to birthdate format validation

Found a wee bug and added a check.

Previously, if the value was coming back like "1909", then the birthdate array wouldn't have a 2nd or 3rd entry, and then calling `birthDateSplit[1].length` would return the error message "can't call length on undefined".

### DIFF
--- a/formSchemas.js
+++ b/formSchemas.js
@@ -88,6 +88,11 @@ const validBirthDateLengths = {
   errorMessage: 'errors.login.dateOfBirth.format',
   validate: value => {
     const birthDateSplit = value.split('/')
+
+    if (birthDateSplit.length !== 3) {
+      return false
+    }
+
     return (
       birthDateSplit[0].length === 4 &&
       birthDateSplit[1].length === 2 &&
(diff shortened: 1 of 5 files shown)


formSchemas.js
```diff
@@ -88,6 +88,11 @@ const validBirthDateLengths = {
   errorMessage: 'errors.login.dateOfBirth.format',
   validate: value => {
     const birthDateSplit = value.split('/')
+
+    if (birthDateSplit.length !== 3) {
+      return false
+    }
```
katedee: ahh thank you 🙇‍♀

pcraig3:
Thanks! Will reconcile this with Nick's PR that just went in and then merge.

PR #41, opened 2019-07-16
Task/Stub out review pages

pcraig3:
## Stub out "review your income" and "review your benefits" pages

Created placeholder pages for these two steps in the flow and linked them together.

For today's demo, I'd like to be able to show a version of the app where we can make it through to the confirmation page, so adding these pages and then changing some of the links lets us do that. (Previously, we would just stop at marital status).

Also including the "do you live in Ontario" page as part of the flow.

## Screenshots

| review your income | review your benefits |
|---------------|-----------------|
| Screen Shot 2019-07-16 at 9 02 05 AM     |    Screen Shot 2019-07-16 at 9 02 15 AM   |

### DIFF
--- a/routes/confirmation/confirmation.controller.js
+++ b/routes/confirmation/confirmation.controller.js
@@ -1,5 +1,8 @@
 module.exports = function(app) {
   app.get('/confirmation', (req, res) =>
-    res.render('confirmation/confirmation', { data: req.session || {} }),
+    res.render('confirmation/confirmation', { data: req.session }),
   )
+
+  // Might make more sense to put this somewhere else?
+  app.get('/review', (req, res) => res.render('confirmation/review', { data: req.session }))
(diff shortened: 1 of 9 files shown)


routes/confirmation/confirmation.controller.js
```diff
@@ -1,5 +1,8 @@
 module.exports = function(app) {
   app.get('/confirmation', (req, res) =>
-    res.render('confirmation/confirmation', { data: req.session || {} }),
+    res.render('confirmation/confirmation', { data: req.session }),
   )
+
+  // Might make more sense to put this somewhere else?
+  app.get('/review', (req, res) => res.render('confirmation/review', { data: req.session }))
```
katedee: I think it's fine here 🤷‍♀ 
That may change later tho

PR #42, opened 2019-07-16
Feature/Add auth page before editing personal information

pcraig3:
## Add another authentication step before changing name or address

One of the things we discussed with the authentication team at CRA was that we would use a layered security approach where we ask people another knowledge based question when they want to make a big change to their info, like changing an address or marital status. 

Adding some middleware that achieves this behaviour.

Expected behaviour

- if someone clicks "change address" or "change marital status", they see the /login/auth page
- they have to give any number, but not random characters and not empty
- once they answer, they can change address / name
- next time they want to edit something, they don't see the screen

Note: since it doesn't matter what their expected benefit is, I don't have any check for a precise number. I just wanted to show a flow in which you have to answer an additional question.

## Use query parameter to redirect

Since the auth page has to redirect to another page that we don't know ahead of time (all of our other redirects are hardcoded), I put the redirect URL as a query parameter. None of our other pages work like this, and at this point, I don't think any of them will have to. The auth page is unique. 

So that we can't be used for phishing people (it's a long shot, but you know), I also added a check that the redirect query parameter must start with a slash (ie, a relative link: `/start` or `/personal/address/edit`), which means we can't use absolute links like `https://google.com` in there.

## gif

auth

### DIFF
--- a/utils/index.js
+++ b/utils/index.js
@@ -49,6 +49,36 @@ const checkPublic = function(req, res, next) {
 }
 
 /**
+ * This request middleware is used to add an "auth" step to some of our pages
+ *
+ * If we are visiting one of our edit pages, we want to ask users
+ * another question before we let them edit their info.
+ *
+ * This means that we want to show the /login/auth page the first time, but not afterwards
+ *
+ * We don't want to show the /login/auth page if
+ *
+ * - we're running tests
+ * - we have already set "auth" to true
+ *
+ * Otherwise, we redirect to the /login/auth page and put the redirect query in the URL
+ */
+const doAuth = function(req, res, next) {
+  // if running tests, do nothing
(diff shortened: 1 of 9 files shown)


utils/index.js
```diff
@@ -48,6 +48,36 @@ const checkPublic = function(req, res, next) {
...
+    return next()
+  }
+
+  // go to original url if "auth" is truthy
+  const { login: { auth = null } = {} } = req.session
+  if (auth) {
+    return next()
+  }
+
+  return res.redirect(`/login/auth?redirect=${encodeURIComponent(req.path)}`)
```
katedee: 😎

pcraig3:
Talked with Nancy and co and we're good to merge.

PR #47, opened 2019-07-16
Task/Set up /login pages for translations

pcraig3:
Add translation methods to all the content in the login pages.

Should be an easy merge.

Also did a couple other things:

- created a mixin for our copy + pasted "Continue / Cancel" buttons
- translated the title in the base.pug template


views/login/code.pug
```diff
@@ -5,29 +5,29 @@ block variables
...
+          <p>#{__('Your code is on the personally-addressed letter.')}</p>
 
       input#redirect(name='redirect', type='hidden', value='/login/sin')
 
       .buttons
         .buttons--left
-          button(type='submit') Continue
+          button(type='submit') #{__('Continue')}
         .buttons--right
           a.button-link.transparent.full-width(href='/clear') Cancel
```
katedee: missed the Cancel here. But also, should we just put these buttons into an include?

pcraig3: I think maybe you're right @katedee. 

We could put in a block that allows you to put in a third option.

I'll get on it.

pcraig3:
let's do it!

PR #51, opened 2019-07-17
Task/Add translations functions to RRSP pages + offramp page

pcraig3:
Like it says on the tin. Updated the RRSP pages and the offramp pages with the translation functionality.

### Rejigged the banner mixin 

We're using a `block` instead of a text thing because I think it's cleaner than writing HTML in as a string. (Sorry @katedee, I didn't notice earlier).

### Using buttons mixins in a few places that we haven't been

Refactored the /deductions pages so that they're using our cool new mixins


views/_includes/banner.pug
```diff
@@ -1,4 +1,7 @@
-mixin banner(text, variant, spanChar)
+mixin banner(variant, spanChar)
   div(class="banner banner--" + variant +"")
     span(class="banner__icon banner--"+ variant +"__icon" aria-hidden="true")= spanChar || 'i'
-    != text
\ No newline at end of file
+    if block
```
sboisvert: Is this supposed to have a name like the block variables or etc or is this some other sort of magic?🔮

pcraig3: it's magic 🐰 

https://pugjs.org/language/mixins.html#mixin-blocks

pcraig3: There's another one here: https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/views/_includes/buttonMixin.pug#L11

PR #52, opened 2019-07-18
Task/Add trillium rental amount page

pcraig3:
This is the first page of the trillium benefits, the rental amount.

Same as the others, I added a currency schema and there's one input on the page. Fairly standard by this point.

Notes:

- I created a tiny function for the isCurrency schema because it's all copy + paste
- I refactored the `*Amount` tests to cut down on all the copy+pasted code in there
- If the updating the `locales` files causes conflicts with anyone else's code, I can remove that from this PR

## Screenshot

Screen Shot 2019-07-18 at 11 38 48 AM


formSchemas.js
```diff
@@ -261,12 +270,7 @@ const rrspSchema = {
 }
 
 const rrspAmountSchema = {
-  rrspAmount: {
-    isCurrency: {
-      errorMessage: 'errors.rrspAmount',
-      options: { allow_negatives: false },
-    },
-  },
+  rrspAmount: currencySchema(),
```
katedee: 👌

PR #53, opened 2019-07-19
Feature/text input include

katedee:
### +textInput
- `label`: text for the label
- `divClasses`: any classes to be applied to the wrapper div (but not the input itself)
- `hintText`: any text that appears above the input in light grey (ex. Optional)
```
+textInput(label, divClasses, hintText)(any attributes you want)
```

with unit example
```
+textInput('Total RRSP contributions')(class='w-1-2 input-with-unit' name='rrspAmount' placeholder='0.00')
```

Now you only have to pass null in the rare case that you want `hintText`, but no `divClasses`


pcraig3:
I don't love all the `null`s, let's see what we can do about them.

Suggestions:

- Mixins can take attributes: https://pugjs.org/language/mixins.html#mixin-attributes
  - Passing in attributes would be better for frequently changing attributes like "class", or rarely used ones like "placeholder"
- There's also the wrapper div in there, so the attributes can go in two places :/ 
  - If it really doesn't work, you can try excluding the wrapper div from the mixin (but it's better to keep it if possible)
- I think we always want to pass "value" in — the fact that it's not in some of our inputs is an oversight
- `unit` is always a `$` right now, and I don't see any designs coming up where it will change: I propose don't even give the option to pass it in
- Also, let's reorder the params by most often used (we always need the key and the label), and then the less commonly used ones, we can just ignore them if we don't need them

Also, what do you think of keeping both inputs mixins in the same file (like with the buttons)?

views/_includes/input-text.pug
```diff
@@ -0,0 +1,31 @@
...
+//- `divClasses`: any classes to be applied to the wrapper div (but not the input itself)
+//- `hintText`: any text that appears above the input in light grey (ex. Optional)
+
+mixin textInput(label, divClasses, hintText)
+  div(
+    class={['has-error']: errors && errors[attributes.name]}
+    class=divClasses
+  )
+    label(
+      for!=attributes.name
```
pcraig3: Does `!=` mean if-exists?

views/_includes/input-text.pug
```diff
@@ -0,0 +1,31 @@
...
+      id!=attributes.name + '__label'
+    ) #{__(label)}
+    if hintText
+      span.cra-form-message #{__(hintText)}
+    if errors && errors[attributes.name]
+      +validationMessage(
+        errors[attributes.name].msg, 
+        attributes.name
+      )
+    if attributes.class.includes('input-with-unit')
```
pcraig3: ooh, clever

views/_includes/input-text.pug
```diff
@@ -0,0 +1,31 @@
...
+      span.input-with-unit__unit(
+        id!=attributes.name+'__unit', 
+        for!=attributes.name
```
pcraig3: we can remove this `for`, since that was actually something I left in by accident. 

I used to have this span as a `<label>` before (so I added a `for` attr) because I saw it here: http://alphagov.github.io/digitalmarketplace-frontend-toolkit/forms/textbox.html#question-100

- adding a label: https://github.com/cds-snc/cra-claim-tax-benefits/commit/1fadc0e708ca3a8fef280bdf16117fa67f9e159a
- changing to a span: https://github.com/cds-snc/cra-claim-tax-benefits/commit/2ebfef80d3ab0d8a1860f5dd3a898d7a6ab47ca2

katedee: ahhh okay. Removed it!

PR #56, opened 2019-07-19
Task/Add checkbox input to the review page

pcraig3:
On the design we have for the review page there's a checkbox input, which we don't have in the app yet.

This PR adds a checkbox to the "/review" stub page while keeping it a stub. That means that the content isn't translated or anywhere near the final version, but you do have to check a checkbox now to continue through to the confirmation page. We're not storing the value in the session so it doesn't add it back in if you reload the page, but I think that's fine at this stage of the game.

Added a POST route as well and a few basic tests for it. 

Changed some of the radios CSS to accommodate for checkboxes & radios, so hopefully that all looks okay.

## gif

review


pcraig3:
Sure, good suggestion.

PR #57, opened 2019-07-19
Task/Add trillium student residence page

pcraig3:
This is the last page of the 5 trillium pages, the student residence page.

It's pretty simple and does the job, but it looks pretty barren and unfortunately our form CSS crunches all the words up on the side of the page.

It's worked so far, but it's not good at all for this page in particular. There's also sometimes that the button text is a bit squished, so I was planning to look at those two things in particular after user testing and see if we can't add a class or two to fix our problems.

@katedee, if you approve then I'll merge and ask Brian on Monday about if we want to fix this before usability testing.

### Screenshots

|  as designed | as built   |
|---|---|
| Screen Shot 2019-07-19 at 6 17 01 PM  |  image |

### DIFF
--- a/formSchemas.js
+++ b/formSchemas.js
@@ -9,6 +9,15 @@ const currencySchema = (errorMessageString = 'errors.currency') => {
   }
 }
 
+const yesNoSchema = (errorMessageString = 'errors.yesNo') => {
+  return {
+    isIn: {
+      errorMessage: errorMessageString,
+      options: [['Yes', 'No']],
+    },
+  }
+}
+
 /**
  * Runs an array of validators over a value
  * this is a workaround I found to allow multiple custom validators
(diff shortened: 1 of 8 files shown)


formSchemas.js
```diff
@@ -9,6 +9,15 @@ const currencySchema = (errorMessageString = 'errors.currency') => {
...
 
+const yesNoSchema = (errorMessageString = 'errors.yesNo') => {
+  return {
+    isIn: {
+      errorMessage: errorMessageString,
+      options: [['Yes', 'No']],
+    },
+  }
+}
+
```
katedee: sweet, when this goes through I'll add it to my PR ✨

routes/deductions/deductions.spec.js
```diff
@@ -83,6 +39,68 @@ describe('Test /deductions responses', () => {
...
+        })
+
+        test('it returns a 422 response for no posted value', async () => {
+          const response = await request(app)
+            .post(yesNoResponse.url)
+            .send({ redirect: '/' })
+          expect(response.statusCode).toBe(422)
+        })
+
+        const badValues = ['', null, false, 0, 'dinosaur', 'yes']
```
katedee: 🦖

PR #59, opened 2019-07-22
🦟 BUG 🐛/Use different node base container in Docker file

pcraig3:
Late Friday afternoon, our builds started magically failing to push to Azure for some reason, and then my manual builds wouldn't run either.

After checking out the application code, it looked like there were no proper changes to the app that would cause a previously-running container to not run anymore. 

Also, creating containers from old commits that ran fine wouldn't run either, so it was a bit of a head scratcher. Not sure exactly the issue we were getting but it was something about the container not starting up on time.

I figured that the thing that might have happened would have been an update to the base container we're relying on.

So after changing from `lts-alpine` to `lts-slim`, the new container runs again on Azure Web Apps. Thank effing goodness.

Also ran a good old npm update just for good measure.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:lts-alpine
+FROM node:lts-slim
 LABEL maintainer="paul.craig@cds-snc.ca"
 
 ARG GITHUB_SHA_ARG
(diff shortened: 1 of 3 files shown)


pcraig3:
🤞

PR #60, opened 2019-07-22
Feature/confirm income

katedee:
### TL;DR:
- new details under "No" button that's very optional
- fake data for confirm income that probably needs to be re-worked
- styled "breakdown" tables
- currency filter to format currencies

I made up some json data for now, and I imagine this won't be how it's structured at all. Open to re-structuring it to something that makes more sense, but at this point re-working the page to pull in different data won't be more than a few minutes 🤞 

Using a mixin block to add the details to No, which only show if "No" is selected. All done with CSS 😎

Added a lil' filter for currency, though we can get rid of it if it turns out we don't need it (if say, the numbers are passed as pre-formatted strings). Set the default local to `en-US` because based on design, this is the correct formatting. 

Once we get this through, I can easily put together [Review and File Tax Return](https://trello.com/c/Uhoe7Tzs/126-create-review-expected-benefit-page-in-html-prototype)

localhost_3005_financial_income (2)


public/scss/_forms.scss
```diff
@@ -169,3 +169,45 @@ form.cra-form {
...
+    font-size: 1.75rem;
+    font-weight: 700;
+  }
+
+  &__subheading {
+    text-align: left;
+    text-decoration: underline;
+    font-weight: 400;
+    padding: 10px 28px;
+    padding-bottom: 20px;
```
pcraig3: Let's use the `$space-md`, etc units where we can for padding and margins.

public/scss/components/_multiple-choice.scss
```diff
@@ -106,3 +106,32 @@
...
+    margin-left: -40px;
+
+    p {
+      margin: 0;
+    }
+
+    &__wrapper {
+      margin-top:40px;
+      padding: 20px;
+      padding-left: 30px;
```
pcraig3: Same again here (`$space-{size}`)

routes/financial/testIncome.json
```diff
@@ -0,0 +1,30 @@
...
+      "name" : "CPP",
+      "amount" : 1213.00
+    },
+    "EI" : {
+      "name" : "Provincial tax",
+      "amount" : 456.00
+    }
+  },
+  "totalTax": 4782.00
+}
```
pcraig3: Do you want to merge this into the `user.json`, or was there a reason not to (other than that I suggested you hardcode it and don't put it in the user.json file)? 👼 

Also, I was taking the "super easy to implement" approach of keeping income as strings so that we don't have to format. (https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/api/user.json#L23). No preference, really.

views/_includes/yesNoRadios.pug
```diff
@@ -11,3 +11,5 @@ mixin yesNoRadios(key, value, question, errors = false)
         .multiple-choice__item
           input(id=`${key}No` name=key type="radio" value="No" checked=(value =='No') aria-describedby=(errors ? `${key}-error` : false))
           label(for=`${key}No`) #{__('No')}
+          if block 
+            block
```
pcraig3: Yeah, this is great.

views/financial/income.pug
```diff
@@ -1,15 +1,69 @@
...
+include ../_includes/yesNoRadios
 
 block variables
   -var title = 'Confirm your income information'
+  -var income = hasData(data, 'financial.income') ? data.financial.income : {}
+  -var taxes = hasData(data, 'financial.taxes') ? data.financial.taxes : {}
+  -var totalIncome = hasData(data, 'financial.totalIncome') ? data.financial.totalIncome : 0
+  -var totalTax = hasData(data, 'financial.totalTax') ? data.financial.totalTax : 0
+  - var count = 1;
+  - var year = new Date().getFullYear()-1
```
pcraig3: You could hardcode this stuff if it were easier, but 🤷‍♂, also not bothered now it's done. 

Although would be better to keep formatting consistent.  (ie, `-var` vs `- var`)

views/financial/income.pug
```diff
@@ -1,15 +1,69 @@
...
 
 block content
 
   h1 #{title}
 
   div
-    p Here's what CRA knows about your income for the current tax year (filing for 2018). Additional claims and deductions can be made afterwards.
+    p #{__(`Here\'s what CRA knows about your income for the current tax year (filing for ${year}). Additional claims and deductions can be made afterwards.`)}
```
pcraig3: Let's use `’` rather than the straight quote.

views/financial/income.pug
```diff
@@ -1,15 +1,69 @@
...
+            
+        tr.breakdown-table__summary-row
+          td #{__(`Total tax paid for ${year}`)}:
+          td $#{currencyFilter(totalTax)}
+
+  form.cra-form(method='post')
+    +yesNoRadios('confirmIncome', null, 'Is this information correct?', errors)
+      .no-info
+        .no-info__wrapper
+          p #{__('If this information doesn’t match your records, you can’t change it here.')}
```
pcraig3: Yeah, here's a curly quote if you wanted to copy+paste.

views/financial/income.pug
```diff
@@ -1,15 +1,69 @@
...
+          p #{__('You should not use this service to file income information you believe to be incorrect.')}
+    
+    input#redirect(name='redirect', type='hidden', value='/financial/income')
```
pcraig3: Is this the right place to redirect to?

views/financial/income.pug
```diff
@@ -1,15 +1,69 @@
...
+          tr
+            td #{__(`${tax.name} deduction`)}: 
+            td $#{currencyFilter(tax.amount)}
+            
+        tr.breakdown-table__summary-row
+          td #{__(`Total tax paid for ${year}`)}:
+          td $#{currencyFilter(totalTax)}
+
+  form.cra-form(method='post')
+    +yesNoRadios('confirmIncome', null, 'Is this information correct?', errors)
```
pcraig3: Interesting 🤔 Hadn't thought about places where there's no value.

pcraig3: (not that anything needs to change. Just interesting to note.)

routes/financial/testIncome.json
pcraig3: I'm not really bothered right now whether we keep this here or not (not really a blocker for testing), but we shouldn't put the benefits review content in a new file too.

I think, at this point, that this could just be part of the user info.

routes/financial/financial.spec.js
```diff
@@ -6,4 +6,27 @@ describe('Test /financial responses', () => {
...
+  })
+
+  test('it returns a 302 and redirects to the same page when YES is selected', async () => {
+    const response = await request(app)
+      .post('/financial/income')
+      .send({ confirmIncome: 'Yes', redirect: '/financial/income' })
+    expect(response.statusCode).toBe(302)
+    expect(response.headers.location).toEqual('/financial/income')
+  })
+})
```
pcraig3: Yeah, that looks like all we need to check, really. 🤓👍

api/user.json
```diff
@@ -20,7 +20,34 @@
...
+      "CPP" : {
+        "name" : "CPP",
+        "amount" : 1213.00
+      },
+      "EI" : {
+        "name" : "Provincial tax",
+        "amount" : 456.00
+      }
+    },
+    "totalTax": 4782.00,
```
pcraig3: Cool, looks good  🤑 🤑

public/scss/components/_multiple-choice.scss
```diff
@@ -106,3 +106,31 @@
     }
   }
 }
+
+.multiple-choice__item {
+  .no-info {
+    max-width: 600px;
+    margin-left: -$multiple-choice-padding-left;
```
pcraig3: Good call 👍

PR #61, opened 2019-07-22
Fixup/Update charity "no" routing + add session variables to deductions fields 

pcraig3:
Pretty small one.

### When you say "no" to donations, don't go straight to the review page 

Let's go see the trillium pages that have been added since this was originally merged.

### Add session variables into RRSP page and donations page

If someone enters an amount, moves to the next page, and then hits "back" in the browser, they should still be able to see their entered amount.

### Changed donations-amount input to mixin

Since #53, we have nice text input mixins, so let's use it.


pcraig3:
lol @katedee ❤️ your reviews!

PR #66, opened 2019-07-22
Add language link + Save language preferences between pages + translate first two pages

pcraig3:
Hello, so this PR is pretty big, but it sets us up to start translating pages pronto.

## Add language link to header

Adds a link to the header based on the way the CRA does it (eg, [Apply for EFILE](https://www.canada.ca/en/revenue-agency/services/e-services/e-services-businesses/efile-electronic-filers/apply-efile.html)) where it floats above the federal identity. 

Works on mobile, works for screenreaders, and it's language-aware (switches based on the current language).

## Add middleware to set a language cookie

We've been able to translate any given page on the fly with query parameters, but previously there was no way to preserve that information through a user's session. 

- Added configuration to our i18n library to set language based on a cookie
- Added middleware to set a cookie if a query param comes in (the same one we already use)

Pretty simple and works for what we need 👌 

## Shim out a new route so that we can give French users a french-language link from the get-go

Created a redirect from `/commencer` to `/start?lang=fr` Doesn't totally solve the problem (ie, our URLs are all still unilingual), but it does mean we can give people a french language site without having to type in a query parameter.

## Translate the first couple pages

Added translations from @ClementineHahn's document for the `/start` page and the `/login/code` page, the first two pages in our app.

## Screenshots

| before | after |
|--------|-------|
|    no "Français" link 😢    |    "Français" link 🎉    |
|   Screen Shot 2019-07-22 at 4 55 26 PM   | Screen Shot 2019-07-22 at 4 55 48 PM   |

### gif

lang


pcraig3:
🇫🇷 🚀

PR #67, opened 2019-07-23
Feature/review benefits

katedee:
### TL;DR:
- `user.json` changed a bunch
- adding `space-row` class to give spacing in a not ideal way (but since we don't have a consistent way of sorting the data right now, it lives like so)
- added a util function to sort the data under projected refund by line number
-moved breakdown-table styling to it's own partial
- spruces up the checkbox styling a little

Honestly, the way I've changed the data is not ideal. But I think when there's more time, and we have a clearer idea of what's going in, we could probably work to group these things appropriately, or create a function that takes it all in, and sorts out into these groups. Open to other ideas!

localhost_3005_review (1)


utils/index.js
```diff
@@ -192,6 +192,16 @@ const currencyFilter = (number, fractionDigits = 2) => {
   })
 }
 
+const sortByLineNumber = (...objToSort) => {
+  const superObj = Object.assign({}, ...objToSort);
+  const filteredObj = Object.entries(superObj).filter(obj => {
+    return typeof obj[1] === 'object' && obj[1] !== null && obj[1] !== undefined && obj[1].hasOwnProperty('line')
```
sboisvert: Maybe it's just me, but I have trouble reading this. Should we add a comment that explains what we're doing (and why) so that in the future we know what it is?

confused

Also, maybe it's just me :)

views/confirmation/review.pug
```diff
@@ -2,19 +2,69 @@ extends ../base
...
+          th.breakdown-table__heading #{__('Your projected refund')}
+          th.breakdown-table__heading +$#{currencyFilter(totalBenefits, 0)}
+      tbody
+        tr
+          th(colspan="2").breakdown-table__subheading #{__('Detailed breakdown of benefits')}
+        each refundLine in refundLines
+          -var lineNumber = refundLine.multiLine ? refundLine.multiLine : refundLine.line
+
+          //- do I like this space-row thing right now? no
+          //- but until our date and how we sort these lines is more solid, it lives
```
sboisvert: thumbs up

views/confirmation/review.pug
```diff
@@ -2,19 +2,69 @@ extends ../base
...
+          tr(
+            class={
+              ['breakdown-table__summary-row space-row']: refundLine.name === "Total Refund",
+              ['space-row']: count === 5 || count === 8 || count === 11 || count === 13
```
sboisvert: What makes 5,8,11,13 special? Should we add a comment to explain?

views/confirmation/review.pug
```diff
@@ -23,7 +73,7 @@ block content
           +validationMessage(errors.review.msg, key)
         .multiple-choice__item
           input(id='review' name='review' type="checkbox" value="review" checked=(value =='review') aria-describedby=(errors ? `review-error` : false))
-          label(for='review') I totally consent to this
+          label(for='review') I confirm that, as far as I know, all the information in this tax return is accurate
```
sboisvert: Does this get translated? It doesn't seem to be part of en.json

views/financial/income.pug
```diff
@@ -23,11 +23,11 @@ block content
         thead
           tr
             th.breakdown-table__heading #{__(`Total income earned in ${year}`)}
-            th.breakdown-table__heading $#{currencyFilter(totalIncome, 0)}
+            th.breakdown-table__heading $#{currencyFilter(totalIncome.amount, 0)}
         tbody
           tr
             th(colspan="2").breakdown-table__subheading #{__(`Detailed breakdown of income earned in ${year}`)}
-          each employer in income
+          each employer in incomeSources
```
sboisvert: I suspect that these won't just be employers. It could be things like OAS, GIC, other provincial programs etc.

api/user.json
```diff
@@ -78,16 +96,53 @@
...
+    "medicalExpenseClaimAmount": {
+      "name": "Medical Expenses",
+      "line": 331,
+      "multiLine": "330/331",
+      "amount": 1227.34
+    },
+    "amountsFromSpouse": {
+      "name": "Amounts transfered from spouse",
+      "line": 326,
+      "amount": 0.00
```
pcraig3: Do we want to show these even if they are `0`?

I understand this will wreak havoc on your `.space-row` logic, but let's double check with Brian what we should do with these in the future.

If it's too much work to actually hide them + keep the grouping in order, we can leave them all in for now.

api/user.json
```diff
@@ -78,16 +96,53 @@
...
+      "amount": 1517.94
+    },
     "singleParentClaim": null,
     "dependentDisabilityClaim": null,
     "medicalExpenseClaim": null,
+    "medicalExpenseClaimAmount": {
+      "name": "Medical Expenses",
+      "line": 331,
+      "multiLine": "330/331",
+      "amount": 1227.34
```
pcraig3: If @sboisvert's #69 PR goes in, you'll have to refactor the logic on his medical expense page and the POST route.

api/user.json
```diff
@@ -98,5 +153,34 @@
...
+    },
+    "Total Refund": {
+      "name": "Total Refund",
+      "line": 484,
+      "amount": 3200.00
+    }
+  },
+  "benefits": {
+    "Canada child benefit": 6639.00,
+    "Climate action incentive": 0.00,
```
pcraig3: Same again here, if @sboisvert's #68 PR goes in, you might have to update the logic here. (Also, I think by default we are giving people Climate Buxx $$$)

locales/en.json
```diff
@@ -207,5 +207,30 @@
...
+  "Disability (316)": "Disability (316)",
+  "Amounts transfered from spouse (326)": "Amounts transfered from spouse (326)",
+  "Medical Expenses (330/331)": "Medical Expenses (330/331)",
+  "Net Federal tax (420)": "Net Federal tax (420)",
+  "Net Provincial tax (428)": "Net Provincial tax (428)",
+  "Total Payable (435)": "Total Payable (435)",
+  "Total Deducted (437)": "Total Deducted (437)",
+  "Total Credits (482)": "Total Credits (482)",
+  "Total Refund (484)": "Total Refund (484)"
+}
```
pcraig3: It might be easier to do a final commit where you delete these new keys, because otherwise it's gonna cause a bunch of conflicts with other PRs.

Might be better to get the code changes in first and then update the locales file in a separate PR.

public/scss/components/_breakdown-table.scss
```diff
@@ -0,0 +1,50 @@
...
+  tr td:last-child,
+  tr th.breakdown-table__heading:last-child  {
+    text-align: right;
+  }
+
+  tr:not(:first-child) {
+    border-left: 2px solid $color-grey-light;
+  }
+
+  .space-row {
```
pcraig3: 🌠 🛶

public/scss/components/_multiple-choice.scss
```diff
@@ -84,15 +84,19 @@
...
   }
 
   > input[type='checkbox'] + label {
+    &::before {
+      top: calc(50% - 20px);
+    }
+
     &::after {
-      top: 11px;
+      top: calc(50% - 20px + 11px);
```
pcraig3: Very lol but the page looks good, so 👍

utils/index.js
```diff
@@ -192,6 +192,16 @@ const currencyFilter = (number, fractionDigits = 2) => {
...
+  })
+  const sortedArrayObj = filteredObj.map(obj => obj[1]).sort((a,b) => a.line - b.line)
```
pcraig3: Nice! So if it hits value that's not an object or doesn't have the "line" key, it will get filtered out, right?

views/confirmation/review.pug
```diff
@@ -2,19 +2,69 @@ extends ../base
 include ../_includes/yesNoRadios
 
 block variables
-  -var title = 'Review your estimated benefits'
+  -var title = 'Review and file tax return'
```
pcraig3: Let's `__()` this bad boy

views/confirmation/review.pug
```diff
@@ -2,19 +2,69 @@ extends ../base
...
 block content
 
   h1 #{title}
 
   div
-    p This is an overview of what you will be submitting to the Canada Revenue Agency (CRA) and all of the benefits you will receive once you have filed.
-    p Make sure that everything looks up-to-date before continuing.
-    p If anything is incorrect, please <a href="#">contact CRA</a>.
+    p This is an overview of your projected tax return, and all the benefits you will receive when you file for the current tax year.
+    p Take one final look before continuing. If something looks incorrect, please #[a(href="https://www.canada.ca/en/revenue-agency/corporate/contact-information/telephone-numbers.html") contact CRA] before filing.
```
pcraig3: Also, let's translate these as well.

views/confirmation/review.pug
pcraig3: LOL!

pcraig3: I love it ❤️ 
We don't have any real way of grouping these, so 🧙‍♀️

views/confirmation/review.pug
pcraig3: Good catch ⚾️

utils/index.js
```diff
@@ -192,6 +192,16 @@ const currencyFilter = (number, fractionDigits = 2) => {
   })
 }
 
+const sortByLineNumber = (...objToSort) => {
```
pcraig3: It would be nice to have 2-3 tests for this but given ⏲ ⌛️ ⌚️, we can do them l8r.

utils/index.js
katedee: yup!

pcraig3:
yolo

PR #69, opened 2019-07-23
Feature/medical claim

sboisvert:
This is really a copy paste job and I think we'll create a better system for beta since it's a bit ridiculous.

accident-band-aid-bandages-42230

PR #70, opened 2019-07-23
Feature/Add in ~70% of French content 

pcraig3:
This adds the content in that big spreadsheet into our locales file.

Since #68 and #69 are pretty much ready to go, I'm happy to wait for those to go in. Maybe even #67 as well.

## Add most french translations

Most of the pages have translations, but some have been missed and a bunch are brand new, so there's no translations for them.

## A bit of cleanup

It also does a few other things:

- takes away the `current tax year (filing for 2018)` sentence construction which seems 1000% more awkward than saying `in 2018`
- translates all the `-var title` strings and not the `title` variables (because sometimes we were double translating)
- removes a few unused error message keys


views/deductions/trillium-rent-amount.pug
```diff
@@ -1,12 +1,12 @@
 extends ../base
 
 block variables
-  -var title = __('Enter rent paid in 2018')
+  -var title = 'Enter rent paid in 2018'
```
katedee: Not concerned about it right now, but going forward we should stick to a consistent way for this

pcraig3: Good point, will fix (and will update the "review" page as well)

PR #74, opened 2019-07-23
Change code length

sboisvert:
Our usability testing has a 9 letter code on the letter, therefore we need to accept 9 characters. We keep 8 characters as also valid to keep the current code working.

### DIFF
--- a/formSchemas.js
+++ b/formSchemas.js
@@ -52,7 +52,7 @@ const loginSchema = {
   code: {
     isLength: {
       errorMessage: 'errors.login.length',
-      options: { min: 8, max: 8 },
+      options: { min: 8, max: 9 },
     },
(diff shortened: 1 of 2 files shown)


pcraig3:
Here's the relevant example: https://github.com/cds-snc/cra-claim-tax-benefits/commit/e5180fca24fd15b88f40ad381ef9a19641880835

formSchemas.js
```diff
@@ -52,7 +52,7 @@ const loginSchema = {
   code: {
     isLength: {
       errorMessage: 'errors.login.length',
-      options: { min: 8, max: 8 },
+      options: { min: 8, max: 9 },
```
pcraig3: I think this should just be `min: 9, max: 9`

PR #75, opened 2019-07-23
Change default access code

sboisvert:
This is a search and replace of QWER1234 to A5G98S4K1.

We are doing this because the code printed on our letter is A5G98S4K1

### DIFF
--- a/routes/login/login.spec.js
+++ b/routes/login/login.spec.js
@@ -80,7 +80,7 @@ describe('Test /login responses', () => {
     expect(response.statusCode).toBe(422)
   })
 
-  const codes = ['QWER1234', 'qwer1234']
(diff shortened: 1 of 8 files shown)


sboisvert:
Should work now

sboisvert:
I'm going to move all changes to this one and close #74

routes/login/login.spec.js
```diff
@@ -80,7 +80,7 @@ describe('Test /login responses', () => {
     expect(response.statusCode).toBe(422)
   })
 
-  const codes = ['QWER1234', 'qwer1234']
+  const codes = ['A5G98S4K1']
```
pcraig3: This should still work with lower-case letters. Had the extra test just to confirm the behaviour.

PR #81, opened 2019-07-23
update for employee names and income amount

katedee:
- add the employee name/display the employee name
- update the values in incomeSources

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -255,5 +255,8 @@
   "Total Deducted (437)": "Total Deducted (437)",
   "Total Credits (482)": "Total Credits (482)",
   "Total Refund (484)": "Total Refund (484)",
-  "I confirm that, as far as I know, all the information in this tax return is accurate": "I confirm that, as far as I know, all the information in this tax return is accurate"
+  "I confirm that, as far as I know, all the information in this tax return is accurate": "I confirm that, as far as I know, all the information in this tax return is accurate",
+  "undefined Total Income": "undefined Total Income",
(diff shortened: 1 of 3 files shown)


locales/en.json
```diff
@@ -255,5 +255,8 @@
   "Total Deducted (437)": "Total Deducted (437)",
   "Total Credits (482)": "Total Credits (482)",
   "Total Refund (484)": "Total Refund (484)",
-  "I confirm that, as far as I know, all the information in this tax return is accurate": "I confirm that, as far as I know, all the information in this tax return is accurate"
-}
+  "I confirm that, as far as I know, all the information in this tax return is accurate": "I confirm that, as far as I know, all the information in this tax return is accurate",
+  "undefined Total Income": "undefined Total Income",
```
pcraig3: This key means that it gets "undefined" at some point. is that happening during a test?

katedee: so when we change/update the data, this'll happen, unless you go back to the start and enter an access code, to force it to pull in fresh user info

views/financial/income.pug
```diff
@@ -29,7 +29,7 @@ block content
             th(colspan="2").breakdown-table__subheading #{__(`Detailed breakdown of income earned in ${year}`)}
           each income in incomeSources
             tr
-              td #{__(`Employer ${count} Total Income`)}
+              td #{__(`${income.name} Total Income`)}
```
pcraig3: Actually, let's only translate the string "Total Income" instead of including the employer name

pcraig3:
One final comment just to note that we are talking about "employer" names.

PR #83, opened 2019-07-23
Fixup/Inputs need labels

pcraig3:
Was doing an axe scan of our existing pages, and I noticed that there were a couple pages where we were getting the error message that labels weren't being associated with inputs.

- the /personal/residence page
- the /personal/address/edit page 

The reason that they weren't being associated was that our `<input>` elements had `name` attributes but no `id`.  

### manually add `id` to /personal/residence `<select>`

Super easy change.

### update the text-input mixin to set the `id` as the `name`

If an `id` is passed in explicitly, it will be used instead of the name value.

## Screenshot

Screen Shot 2019-07-23 at 3 09 04 PM

### DIFF
--- a/views/_includes/input-text.pug
+++ b/views/_includes/input-text.pug
@@ -8,14 +8,14 @@ mixin textInput(label, divClasses, hintText)
     class=divClasses
   )
     label(
-      for!=attributes.name 
+      for!=attributes.name
       id!=attributes.name + '__label'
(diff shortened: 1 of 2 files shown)


pcraig3:
Not sure why the test is failing. I think it's a github actions thing. Will wait and see if it comes back in 30 minutes.

PR #84, opened 2019-07-23
update values on review page

katedee:
Updating the values as requested, except for RRSP which is being handled separately and is not guaranteed for review.

localhost_3005_review (2)

PR #85, opened 2019-07-23
Change RRSP amount to new style

sboisvert:
Previously we only had one property called rrspAmount that had an integer as value.
We now have the *Amount properties contain an object that has the name, the line number, and the amount.

This PR has #84 merged in already

### DIFF
--- a/api/user.json
+++ b/api/user.json
@@ -96,6 +96,11 @@
       "amount": 0.00
     },
     "rrspClaim": null,
+    "rrspAmount": {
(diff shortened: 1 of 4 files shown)


sboisvert:
@pcraig3 Now with en.json reverted

katedee:
one thing! could we change `space-row` in the income.pug file to be 5,6,7,9 and follow the design (now that we have RRSP)

pcraig3:
Sure, I'll do it.

PR #88, opened 2019-07-24
Alpha Banner

nmakuch:
### 1) Creating an Alpha Banner component that is a part of the `base.pug` layout

| Desktop | 
|---------|
| Screen Shot 2019-07-24 at 13 57 49    | 

| Mobile |
|--------|
| Screen Shot 2019-07-24 at 13 58 18    |

#### Breakdown
- Designing alpha banner component
- Converting design to html 
- Mobile queries
- French translations
- Css styling

### 2) Aligning french toggle with fip

| Before | After |
|--------|-------|
|   Screen Shot 2019-07-24 at 14 03 01     |   Screen Shot 2019-07-24 at 14 02 17    |

#### Breakdown
- Grouping fip and french toggle into same container
- Applying flexbox properties for alignment
- Mobile queries (`flex: row` on desktop >>> `flex:column` on mobile)

### DIFF
--- a/public/scss/_base.scss
+++ b/public/scss/_base.scss
@@ -76,6 +76,50 @@ li:not(:last-of-type) {
   @extend %page-container;
 }
 
+.fip-container {
+  flex-direction: row;
+  justify-content: space-between;
+  display: flex;
+
+  @include sm {
+    justify-content: space-between;
+  }
+}
+
+.phase-banner {
+  background: #F4F4F4;
+}
+
+.phase-banner {
+  div {
+  @extend %page-container;
+  display: flex;
+  align-items: center;
+  padding-top: $space-sm;
+  padding-bottom: $space-sm;
+  margin-bottom: $space-md;
+  }
+
+  span {
+    font-size: 0.5em;
+
+    @include sm {
+      font-size: 0.7em;
+    }
+  }
+
+  span:first-of-type {
+    border: 2px $color-black solid;
+    padding: 1px 8px;
+    letter-spacing: 1px;
+    margin-right: $space-md;
+  }
+
+  span a {
+    color: $color-black;
+  }
+}
+
 .pure-table {
   width: 100%;
 
--- a/public/scss/_base.scss
+++ b/public/scss/_base.scss
@@ -80,7 +80,7 @@ li:not(:last-of-type) {
   flex-direction: column;
   display: flex;
 
-  @include xs {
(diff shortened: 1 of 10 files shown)


views/base.pug
```diff
@@ -28,15 +28,21 @@ html(lang=getLocale() === 'fr' ? 'fr' : 'en')
   body
     .container
       header
+        .phase-banner
+          div
+            span ALPHA
+            span This is a new service – your 
+              a(href='#') #{__('feedback')}
```
pcraig3: If we're putting in a link, we should actually have it go somewhere. (maybe a google form?)

nmakuch: removed link and updated to "This site will change as we test ideas."

public/scss/components/_header.scss
```diff
@@ -1,14 +1,19 @@
...
-  padding: $space-xs 0 0 0;
+  padding: 0 0 0 0;
+
+  button {
+    padding: 0;
+  }
 
   @include sm {
-    padding: $space-lg 0 $space-xl 0;
+    padding: 0 0 $space-xl 0;
```
pcraig3: This could be simplified to `padding-bottom: space xl`

public/scss/_base.scss
```diff
@@ -76,6 +76,62 @@ li:not(:last-of-type) {
...
+  display: flex;
+
+  @include sm {
+    justify-content: space-between;
+    flex-direction: row;
+  }
+}
+
+.phase-banner {
+  background: #F4F4F4;
```
pcraig3: Would one of our other colours work? If not, it's fine. If it's just in this one place, we're probably okay.

public/scss/_base.scss
```diff
@@ -76,6 +76,62 @@ li:not(:last-of-type) {
   @extend %page-container;
 }
 
+.fip-container {
```
pcraig3: Actually, can you make this an extra file? We should try to keep base as small as reasonable.

public/styles.scss
```diff
@@ -20,3 +20,5 @@
 @import './scss/components/_multiple-choice';
 @import './scss/components/_banners';
 @import './scss/components/_breakdown-table';
+@import './scss/components/_fip';
+@import './scss/components/_phase-banner';
```
pcraig3: Could have put them both in one file if you wanted, but this works too 👍

PR #90, opened 2019-07-24
[DO NOT MERGE] Change hint texts for login flow

pcraig3:
⚠️ **DO NOT MERGE** ⚠️ 

***

Hillary wants the example hint text changed for the login flow so that testing will look more realistic.

Because we want to demo the app internally, we want the hint text to work for people trying it out so that they don't get locked out if they just want to take a look.

Once the user testing is over, I'll close this PR.


pcraig3:
Closing because we needed this for usability testing and now it's done.

PR #101, opened 2019-07-26
Add autofocus to form inputs

katedee:
From the usability feedback, there was some issue with the inputs not having focus. People would type what was on the letter provided, but look up, and see a blank input and be confused.

To solve this, let's autofocus on the first input, or on errors, the first input in the list of errors.

This is a start, as I haven't tackled all the forms we have in the app.

WIP because I need to add tests

### DIFF
--- a/views/login/code.pug
+++ b/views/login/code.pug
@@ -12,7 +12,7 @@ block content
 
   div
     form.cra-form(method='post')
-      +textInput('Personal access code', null, 'For Example: A5G98S4K1')(class='w-3-4' id='code' name='code', autofocus, value=data.code)
+      +textInput('Personal access code', null, 'For Example: A5G98S4K1')(class='w-3-4', id='code' name='code', autofocus, value=data.code)
 
       div
         details.
--- a/views/login/code.pug
+++ b/views/login/code.pug
@@ -17,7 +17,7 @@ block content
         span.cra-form-message #{__('For Example:')} A5G98S4K1
         if errors
           +validationMessage(errors.code.msg, 'code')
-        input.w-3-4#code(name='code', type='text', value=data.code, autocomplete='off' aria-describedby=(errors && errors.code ? 'code-error' : false))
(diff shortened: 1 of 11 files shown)


views/login/code.pug
```diff
@@ -17,7 +17,7 @@ block content
         span.cra-form-message #{__('For Example:')} A5G98S4K1
         if errors
           +validationMessage(errors.code.msg, 'code')
-        input.w-3-4#code(name='code', type='text', value=data.code, autocomplete='off' aria-describedby=(errors && errors.code ? 'code-error' : false))
+        input.w-3-4#code(name='code', type='text', autofocus, value=data.code, autocomplete='off' aria-describedby=(errors && errors.code ? 'code-error' : false))
```
pcraig3: Could we use the text-input mixin here instead of keeping it as-is?

views/personal/address-edit.pug
```diff
@@ -21,7 +21,7 @@ block content
       h2 #{__('Your mailing address')}
 
       div
-        +textInput('Street number', ['w-1-4','d--ib'])(class='w-3-4', value=streetNumber, name='streetNumber')
+        +textInput('Street number', ['w-1-4','d--ib'])(class='w-3-4', value=streetNumber, name='streetNumber' autofocus=(!errors || errors && firstError.param === 'streetNumber' ? true : false))
```
pcraig3: Can we put the `firstError` var in this template itself since it's the only time we use it?

katedee: definitely could. I was thinking about in the future when there might be more forms with multiple inputs. But you're right that for now it's just this page

views/login/code.pug
katedee: sure thing

views/personal/address-edit.pug
pcraig3: I think maybe you deleted 'firstError' 🤔

PR #103, opened 2019-07-29
add banners to main page and sin page

katedee:
Add the previous banner to the other spots where it's present in the app (according to the latest sketch).
Had to add a check variant, for the one's where the circle contains a check mark inside.

localhost_3005_start
localhost_3005_login_sin

### DIFF
--- a/public/scss/components/_banners.scss
+++ b/public/scss/components/_banners.scss
@@ -30,7 +30,7 @@
         left: 20px;
     }
 
-    &--blue {
+    &--blue, &--check {
         background: $color-banner-blue-light;
         border-left: 5px solid $color-banner-blue;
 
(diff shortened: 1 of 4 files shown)


public/scss/components/_banners.scss
```diff
@@ -30,12 +30,23 @@
...
 
         &__icon {
             background: $color-banner-blue;
         }
     }
+
+    &--check {
+        &__icon {
+            display: inline-block;
+            background-image: url("data:image/svg+xml,%3Csvg width='13px' height='11px' viewBox='0 0 13 11' version='1.1' xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink'%3E%3Cdesc%3Echeckmark%3C/desc%3E%3Cg id='Page-1' stroke='none' stroke-width='1' fill='none' fill-rule='evenodd' stroke-linecap='square'%3E%3Cg id='1.3-Enter-SIN' transform='translate(-79.000000, -232.000000)' stroke='%23FFFFFF' stroke-width='2'%3E%3Cg id='Group-3' transform='translate(81.000000, 232.000000)'%3E%3Cpath d='M0,6.44810072 L3.16500985,8.84667626' id='Line-5'%3E%3C/path%3E%3Cpath d='M3.03186405,1.92361281 L9.77998686,8.5' id='Line-5' transform='translate(6.279987, 5.000000) rotate(84.000000) translate(-6.279987, -5.000000) '%3E%3C/path%3E%3C/g%3E%3C/g%3E%3C/g%3E%3C/svg%3E");
```
pcraig3: awesome, this looks great!

views/_includes/banner.pug
```diff
@@ -1,6 +1,6 @@
-mixin banner(variant, spanChar)
+mixin banner(variant, spanChar='i')
```
pcraig3: Yeah, seems better.

views/login/sin.pug
```diff
@@ -8,6 +9,9 @@ block content
 
   h1 #{title}
 
+  +banner('check', '')
```
pcraig3: Can you pass `null` in instead? Otherwise it looks like the empty space is doing something.

views/_includes/banner.pug
pcraig3: Actually, so far we're only passing in `i` or nothing, so we could just not have a second parameter and write an 'if' based on the string that gets passed in.

But either works, really.

PR #108, opened 2019-07-31
Task/Fix <details> element on IE11

pcraig3:
Copy+pasted in a polyfill into our /public folder and conditionally including it on pages with a `details` element in older browsers (eg, in IE11).

So there's a few decisions in here.

## Adding a polyfill

`<details>` elements don't work in IE11, but they're a nice semantic solution for what we want to do: conditionally display information.

The easy way to fix this without a lot of finagling is to include a polyfill. I've picked the `details-element-polyfill` on npm because it's the most popular one.

## Not using a CDN

After humming and hawing, using a CDN might cause weird problems because @ayoajila has been telling me there are weird proxy issues going on trying to run the app on his laptop.

Using a CDN also means that our container isn't self-contained anymore if it relies on an external request.

So, I just copied the polyfill lock, stock into a `/js` directory. I doubt it gets updated very often and if we need to update it we can just copy in the latest version.

I figured we don't really need to include this in our package.json as a dependency. (what do you think @katedee?)

## Using a conditional statement with `document.write()`

There are more ways to do this that are discussed at length in this this other article: https://philipwalton.com/articles/loading-polyfills-only-when-needed/

However, this was cheap, easy, and it works, so I didn't want to overdo it.

Source:
- https://stackoverflow.com/questions/45509701/elegant-way-to-include-es6-promise-for-ie

## Screenshots

| before | after |
|--------|-------|
|   ie114   | ie113  |

### DIFF
--- a/public/js/details-element-polyfill.js
+++ b/public/js/details-element-polyfill.js
@@ -1,0 +1,203 @@
+/*
+Details Element Polyfill 2.3.1
+Copyright © 2019 Javan Makhmali
+ */
+(function() {
+  "use strict";
+  var element = document.createElement("details");
+  element.innerHTML = "<summary>a</summary>b";
+  element.setAttribute("style", "position: absolute; left: -9999px");
+  var support = {
+    open: "open" in element && elementExpands(),
+    toggle: "ontoggle" in element
+  };
+  function elementExpands() {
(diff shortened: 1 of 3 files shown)


pcraig3:
I've watched the repo, so if anything changes, I'll get an email.

PR #111, opened 2019-08-01
update address to conform to data shape requested

katedee:
mainly that address is now line1 for street address
and line2 for apartment number
I've updated to calling it line1 + line2 in most places, despite it not being the clearest name for it. But otherwise we will get into situations where here you refer to it one way, and here your refer to it another way, and I'd rather keep it consistent for simplicity's sake. Decision not set in stone, though

localhost_3005_personal_address_edit

### DIFF
--- a/formSchemas.js
+++ b/formSchemas.js
@@ -218,7 +218,7 @@ const maritalStatusSchema = {
 const addressSchema = {
   line1: {
     isEmpty: {
-      errorMessage: 'errors.address.streetName.empty',
+      errorMessage: 'errors.address.line1.empty',
       negated: true,
     },
   },
--- a/formSchemas.js
+++ b/formSchemas.js
@@ -216,7 +216,7 @@ const maritalStatusSchema = {
 }
 
 const addressSchema = {
-  streetName: {
+  line1: {
     isEmpty: {
       errorMessage: 'errors.address.streetName.empty',
       negated: true,
--- a/locales/en.json
+++ b/locales/en.json
@@ -13,7 +13,7 @@
   "errors.login.dateOfBirth.validMonth": "Please enter a valid month between 01 and 12",
   "errors.login.dateOfBirth.validDay": "Please enter a valid day",
   "errors.login.dateOfBirth.validYear": "Please enter a valid year",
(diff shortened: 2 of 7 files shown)


formSchemas.js
```diff
@@ -216,7 +216,7 @@ const maritalStatusSchema = {
 }
 
 const addressSchema = {
-  streetName: {
+  line1: {
     isEmpty: {
       errorMessage: 'errors.address.streetName.empty',
```
pcraig3: Can you change this to `errors.address.line1.empty` just so it all matches up (and then update the locales files)?

views/personal/address-edit.pug
```diff
@@ -21,10 +20,7 @@ block content
     form.cra-form.cra-form--lg(method='post')
       h2 #{__('Your mailing address')}
 
-      div
-        +textInput('Street number', ['w-1-4','d--ib'])(class='w-3-4', value=streetNumber, name='streetNumber' autofocus=(!errors || errors && firstError.param === 'streetNumber' ? true : false))
-
-        +textInput('Street name', ['w-3-4','d--ib'])(class='w-3-4', value=streetName, name='streetName')
+      +textInput('Street address', ['w-3-4','d--ib'])(class='w-3-4', value=streetAddress, name='line1' autofocus=(!errors || errors && firstError.param === 'line1' ? true : false))
 
       +textInput('Apartment number', 'w-1-2', 'Optional')(class='w-3-4', value=aptNumber name='aptNumber')
```
pcraig3: You need to change the `name` of this input as well otherwise submitting a new apartment number doesn't update.

locales/en.json
```diff
@@ -13,7 +13,7 @@
   "errors.login.dateOfBirth.validMonth": "Please enter a valid month between 01 and 12",
   "errors.login.dateOfBirth.validDay": "Please enter a valid day",
   "errors.login.dateOfBirth.validYear": "Please enter a valid year",
-  "errors.address.streetName.empty": "Street name can’t be empty",
+  "errors.address.line1.empty": "Street address can’t be empty",
```
pcraig3: there's a line like this in the `fr.json` file as well 😉

PR #112, opened 2019-08-01
Fix/Allow french currency separator

pcraig3:
## Allow french formatted decimal separator 

English formatting looks like `$10.00`, whereas French formatting looks like `$10,00`. This PR makes it possible to submit values like:

- `$10,00`
- `10,00`
- `,10`

into our currency inputs.

Not sure if this feature is really needed because we haven't ever seen anyone stumble over this, so I can kill the PR if we would rather take a "wait and see" approach.


pcraig3:
Gonna close because there's no clear need. Thanks @brianhendrick1.

PR #114, opened 2019-08-02
Create "doRedirect" middleware + remove "validateRedirect" middleware

pcraig3:
All of our "redirect" code uses the passed-in redirect body param if everything else passes.
Created a "redirect" middleware function to abstract this stuff. 

Removed the `validateRedirect` middleware at the start and added a `doRedirect` middleware at the end that will actually run the redirect. Also means we can get rid of many of our post functions.

Pros:

- Save some lines of code
- Enforced consistency of redirect code
- Don't need separate POST functions when the post methods are just updating a session value

Cons:

- Less straightforward because you don't actually see the redirect happen
- The redirect validation only happens at the end rather than be beginning -- could be annoying to change all the tests
- Not sure that it makes the code easier to understand

***

- Also added a few more 200 tests for our offramp pages
- Added a truthy check for `value` in our birth date validation methods


utils/index.js
```diff
@@ -144,6 +144,18 @@ const validateRedirect = (req, res, next) => {
   return next()
 }
 
+//POST functions that handle setting the login data in the session and handle redirecting to the next page or sending an error to the client.
+//Note that this is not the only error validation, see routes defined above.
+const doRedirect = (req, res) => {
```
katedee: 👌 ✨

pcraig3:
Okay @katedee, you allowed this to happen.

PR #115, opened 2019-08-02
Blank is valid answer to currency questions

sboisvert:
During usability testing some users had trouble with pressing "continue" because the placeholder was showing 0. Having a default value of 0 also made the users type in numbers incorrectly.
This patch changes it so that all fields that follow currencySchema can accept blank (and null).
Since we no longer need to have the logic separately for political contributions we just move them to use the currency schema.
For the /auth/ page the test have been changed to accept an empty value since it could be empty.

### DIFF
--- a/routes/login/login.spec.js
+++ b/routes/login/login.spec.js
@@ -372,12 +372,12 @@ describe('Test /login responses', () => {
       expect(response.statusCode).toBe(200)
     })
 
-    test('it returns a 422 response for no posted value', async () => {
+    test('it returns a 302 response for no posted value', async () => { // since 0 is what is displayed by default, a user could leave it blank and press enter. This might not be great in user testing scenario but should be a valid answer
       const response = await request(app).post('/login/auth')
-      expect(response.statusCode).toBe(422)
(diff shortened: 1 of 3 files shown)


routes/login/login.spec.js
```diff
@@ -372,12 +372,12 @@ describe('Test /login responses', () => {
       expect(response.statusCode).toBe(200)
     })
 
-    test('it returns a 422 response for no posted value', async () => {
+    test('it returns a 302 response for no posted value', async () => { // since 0 is what is displayed by default, a user could leave it blank and press enter. This might not be great in user testing scenario but should be a valid answer
       const response = await request(app).post('/login/auth')
-      expect(response.statusCode).toBe(422)
+      expect(response.statusCode).toBe(302)
```
pcraig3: Interesting, this means that if the key is missing, it still works. 🤔 

Doesn't really matter for now, but something to keep in mind.

sboisvert: Ya, it'll work because 0 is a valid response AND we don't have any validation on that value. When we add validation empty will mean 0 which will most likely not be the correct value.

PR #117, opened 2019-08-05
Make the no info-box on yes/no radios more accessible

katedee:
- added some js to toggle aria-expanded on yes/no radios, but only if they have the no-info div for the no option
- added visibility visible/hidden css to boot (just to be sure screen readers aren't reading it)
- checked with axe in chrome 👍

### DIFF
--- a/views/_includes/script.pug
+++ b/views/_includes/script.pug
@@ -7,5 +7,25 @@ script.
   }
 
 script.
+  var noInfoBox = document.getElementsByClassName('no-info')[0]
+
+  if (noInfoBox) {
+    var yesNoRadios = document.querySelectorAll('div.multiple-choice__item')
+    
+    var noInput = yesNoRadios[1].childNodes[0]
+
+    for (var i = 0, len = yesNoRadios.length; i < len; i++) {
+      yesNoRadios[i].addEventListener('click', function(e) {
+        if(e.target.value === 'No') {
+          noInput.setAttribute('aria-expanded', 'true');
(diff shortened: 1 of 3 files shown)


views/_includes/script.pug
```diff
@@ -6,6 +6,26 @@ script.
     })
   }
 
+script.
+  var noInfoBox = document.getElementsByClassName('no-info')[0]
+
+  if (noInfoBox) {
+    var yesNoRadios = document.querySelectorAll('div.multiple-choice__item')
+    
+    var noInput = yesNoRadios[1].childNodes[0]
```
pcraig3: Wouldn't work if there are 2 on a page, but let's worry about it when it happens.

views/_includes/script.pug
```diff
@@ -6,6 +6,26 @@ script.
...
+
+    for (var i = 0, len = yesNoRadios.length; i < len; i++) {
+      yesNoRadios[i].addEventListener('click', function(e) {
+        if(e.target.value === 'No') {
+          noInput.setAttribute('aria-expanded', 'true');
+        } else {
+          noInput.setAttribute('aria-expanded', 'false');
```
pcraig3: I thought this might go on the `div.no-info`, but you're right. 

👍

views/_includes/script.pug
katedee: yeah, I worried about that too, but I guess we can cross that bridge when we get there

PR #118, opened 2019-08-05
split up giant formSchemas file

katedee:
RIP giant `formSchemas.js` file
This PR splis that giant file up into multiple files, basing it off of our routes, and a `utils.schema.js` file which contains the common functions used in our schemas.

On one hand, splitting them out into multiple files could make it easier to delete and parse parts of the app. 

On the other hand, it could be a more confusing set up to understand.

I seem to be having issues with test randomly failing, so there's that too (happened once or twice, and upon re-run no issues)


pcraig3:
I am chill with this going in. 

Makes it easier to test, should we need to.

PR #120, opened 2019-08-09
Task/Add Secure Content Policy (CSP)

pcraig3:
Added a content security policy. 

Docs are here: https://helmetjs.github.io/docs/csp/

Note that if we're including 3rd party scripts or images at any point (google analytics, etc), they will not run unless we change this. You'll get a console error about it, so it will show up as a problem, but just so you know.

Also, I moved our inline JS into a separate file. here's the reasoning behind that:

## Move our JS code into a 'script' file 

This will make our Content Security Policy happier.

The idea is that if you allow inline scripts and if someone can cross-site script you to inject HTML, then they can just write JS straight into your page and someone else's browser will execute it.

If you ban inline <script> execution, then someone will have to inject a script tag pointing at a file in your domain, which is way harder to do.


public/js/script.js
```diff
@@ -0,0 +1,34 @@
...
+var elements = document.querySelectorAll('a.button-link')
+for (var i = 0, len = elements.length; i < len; i++) {
+  elements[i].addEventListener('keydown', function(e) {
+    if (e.keyCode == 32) {
+      e.target.click()
+    }
+  })
+}
+
+// Add aria labels to the revealing/concealing content underneath "yes/no" radios
```
katedee: thanks for adding the comments 👌

PR #134, opened 2019-09-09
Feature/back button

katedee:
We have a back button!

Routing mostly stolen from [Tim's handy routing functions](https://github.com/cds-snc/node-starter-app/blob/master/utils/route.helpers.js)

I initially overcomplicated it, thinking we would navigate back to edit pages (ex edit address), but then remembered way to late that we just go back to the default confirm page when done editing. Still, the work is handy for if you answered "yes" to some of our questions, and then entered an amount. Happy to rip that all out if we're not finding it useful though.

Because of the above work, on edit/amount pages, there's an extra property to the routes that is "editInfo", which just points to where it needs to check for this information (or you can pass it a non-existing property, like `skip`, to have it skip the page in the flow)

I made an assumption that we want a back button on offramp pages, but we might not. I'd also like to handle it differently than I am right now but would like to address that separately

I'd also like to split up our utils into smaller files, but I figured I would handle that separately

image


config/routes.config.js
```diff
@@ -0,0 +1,35 @@
...
+  { name: 'donations amount', path: '/deductions/donations/amount', editInfo:  'deductions.charitableDonationAmount' },
+  { name: 'trillium rent', path: '/trillium/rent/amount', editInfo: 'deductions.trilliumRentAmount' },
+  { name: 'trillium property tax', path: '/deductions/trillium/propertyTax/amount', editInfo: 'deductions.trilliumPropertyTaxAmount' },
+  { name: 'trillium student residence', path: '/trillium/studentResidence'},
+  { name: 'trillium energy amount', path: '/trillium/energy/amount', editInfo: 'deductions.trilliumEnergyAmount' },
+  { name: 'trillium long term care amount', path: '/trillium/longTermCare/amount', editInfo: 'deductions.trilliumLongTermCareAmount' },
+  { name: 'climate action incentive', path: '/deductions/climate-action-incentive' },
+  { name: 'review', path: '/review' },
+  { name: 'confirmation', path: '/confirmation' },
+]
```
pcraig3: I have a potential refactor that breaks compatibility with Tim's code, but hear me out. 

Our "route" dictionaries have two keys: `name` and `path`, both of which are unique. As in, for any given route, the `name` is unique and so is the `path`. 

So finding a route by `name` is the same as finding a route by `path`. 

Therefore, using a name creates a new id-type value that that we didn't have before and is _only_ used as a lookup value. 

SO, (galaxy brain), since we always call `getPreviousRoute` inside of middleware, which has access to the request  object (`req`), which always knows about the url path `req.path`, we shouldn't need to pass another value ever to `renderWithData`. We can just let it figure out the link itself using the request object.

We could potentially just call `getPreviousData` with `req.path`, and then refactor `getRouteWithIndexByName` to find by path rather than name. 

That way

- we don't need the second param in `renderWithData`
- we don't need the `name` key in our routes

Does that all hang together or have I overlooked something?

utils/index.js
```diff
@@ -187,7 +191,13 @@ const hasData = (obj, key) => {
...
     }
     obj = obj[x]
-    return true
+
+    if(!returnVal) {
+      return true
+    } 
+
+    return obj
+
```
pcraig3: Let's write a test for this.

utils/index.js
```diff
@@ -221,6 +231,70 @@ const sortByLineNumber = (...objToSort) => {
...
+ * @returns { name: "", path: "" }
+ */
+const getPreviousRoute = (name, session, routes = defaultRoutes) => {
+  const route = getRouteWithIndexByName(name, routes);
+
+  if (!route || (!"index" in route && process.env.NODE_ENV !== "production")) {
+    throw new Error(
+      "Previous route error.  \n Did you forget to pass it along to renderWithData? \n i.e. renderWithData('login/sin', 'login sin')"
+    );
+  }
```
pcraig3: This might change in light of my earlier suggestion ☝️

utils/index.js
```diff
@@ -221,6 +231,70 @@ const sortByLineNumber = (...objToSort) => {
...
+    ) {
+        // if they didn't do any editing, skip over the edit page
+        routeIndexBack = 2
+      }
+    
+    return routes[Number(route.index) - routeIndexBack] ? routes[Number(route.index) - routeIndexBack] : false;
+  }
+
+  if (!prevRoute()) {
+    return DefaultRouteObj;
```
pcraig3: Does this ever happen?

views/_includes/backButton.pug
```diff
@@ -0,0 +1,2 @@
+if prevRoute
+  a(class='back-button' href=prevRoute.path)= '< Go back'
```
pcraig3: Not to critique Nick's design but the "less than" symbol looks kind of unfinished.

The GOV.UK back link uses a nice little triangle, but this story isn't really so much about the visual design of this button.

Let's use a unicode arrow in the meantime (←) and wrap it in an aria-hidden span. 

```html
<a><span aria-hidden="true">← </span>Go back</a>
```

views/confirmation/confirmation.pug
```diff
@@ -7,6 +7,8 @@ block content
 
   h1 #{title}
 
+  include ../_includes/backButton
+
```
pcraig3: I don't think we need to now (or maybe ever, but it looks like the following code snippet is copy+pasted on all o four pages.

```
  h1 #{title}

  include ../_includes/backButton
```

There's maybe an easy way to refactor this, but for right now let's not since it won't really give us that much.

config/routes.config.js
katedee: No that checks out and I was thinking about that too. I'll work on it in the AM

utils/index.spec.js
```diff
@@ -51,6 +59,33 @@ describe('Test hasData function', () => {
   test('returns true for disabilityClaim', () => {
     expect(hasData(user, 'deductions.disabilityClaim')).toBe(false)
   })
```
pcraig3: Let's write a test for `hasData` so that we see it returns the value, not just true or false.

pcraig3: Oh wait, you removed the logic. Okay, I'm getting rid of the extra param in hasData.

PR #139, opened 2019-09-13
Task/Use dot indexes for all the text on the start page

pcraig3:
We're thinking of moving all the text to the JSON files under `/locales`, but it would be a lot of work and we're not sure we need to do all that stuff yet.

As a small test, I have changed the start page to use indexed values instead of putting all the english text right in the template.


locales/en.json
```diff
@@ -25,19 +25,23 @@
...
+  "service.name": "Claim Tax Benefits",
+  "start.banner": "Keep your notification letter with you. You need information on that letter to use this service.",
+  "start.p1.1": "Welcome to the Canada Revenue Agency’s new",
+  "start.p1.2": "service",
+  "start.p2.1": "This service is to claim tax benefits online.",
+  "start.p2.2": "You are eligible for this service if:",
+  "start.li1": "You earned under $35,000 in 2018 (from employment, pension, RRSP withdrawals, or from benefits and credits you received)",
+  "start.li2": "You do not have self-employment income to report",
+  "start.li3": "You do not earn investment income (except for nominal interest income from Canadian sources)",
+  "start.p3": "In order to claim tax benefits online, you must have the following:",
```
katedee: even though it's just the one paragraph, do we want to just keep it consistent with `start.p3.1`?

katedee:
This seems fine by me, I guess the question is will others actually use it to update the content? Let's check and see if so before we go converting all of it.
But otherwise, Looks Good 2 Me

pcraig3:
🤙

PR #142, opened 2019-09-17
Task/Add "yes/no" page for the property tax + energy costs trillium benefits

pcraig3:
This PR does a bunch of things:

### Adds yes/no pages for property tax and energy costs on a reserve

Now you have to say that you have these costs before entering the amount. Once these are in, there is one page left to go. (Also adds unit tests + updates cypress tests)

### Returned stored `true`/`false` values to the "Yes/No" field for the trillium questions

As mentioned [in this story on trello](https://trello.com/c/BsP43ovG/229-%F0%9F%95%B7bug-a-bunch-of-pages-in-the-app-dont-preserve-users-previously-entered-values), lots of pages don't fill the "yes/no" back when you revisit them. It turns out this is because we have "true/false/null" values and we need to explicitly change them into "Yes" or "No".

So I updated them for the trillium questions but not the rest of the app.

Commit: https://github.com/cds-snc/cra-claim-tax-benefits/commit/6809d8ad041c8305eb8f1d7bd4b4545245ab6819

### Don't put a "0" in the trillium amount fields by default

We heard this was a problem during the usability study and it makes more sense to remove the zeros and just show the placeholder text (`0.00`). Since the default value is `0`, I did an explicit check before putting the value into the mixin.

Commit: https://github.com/cds-snc/cra-claim-tax-benefits/commit/b99295dba19678bc16fe6f83b9b70a4fe20e8c4e

### Adds middleware for yes/no questions

After copy+pasting a bunch of similar-looking messages, it seemed like we could write another ✨middleware ✨ to help us out.

[Wrote a function](https://github.com/cds-snc/cra-claim-tax-benefits/pull/142/commits/bae0a9d60364805368ce156c0dbc40501f2e2e7e) that generalises the common pattern, and then [used it for the trillium questions here](https://github.com/cds-snc/cra-claim-tax-benefits/commit/f4fe1b621ef6ac6b4513ad9a1b264c3f2dbb98f2).

@katedee, I was planning to move the charity question etc over to use this middleware (have to change some variable names), but wanted you to have a look before changing a bunch of existing code. I changed the RRSP page question [here](https://github.com/cds-snc/cra-claim-tax-benefits/pull/142/commits/ff24f731e8653c82e68a2fd25deb79cbfeb767aa). See what you think.

### TODO

- Add one more trillium question
- (opt) Get the other `/deduction` yes/no pages using my middleware

#### Property tax pages

| before | after |
|--------|-------|
|   N/A     |  Screen Shot 2019-09-17 at 12 32 36 PM   |
|  Screen Shot 2019-09-17 at 12 32 55 PM  |   Screen Shot 2019-09-17 at 12 32 45 PM   |

#### Home energy on a reserve pages
| before | after |
|--------|-------|
|   N/A     |  Screen Shot 2019-09-17 at 12 33 33 PM  |
|   Screen Shot 2019-09-17 at 12 33 18 PM   |  Screen Shot 2019-09-17 at 12 38 45 PM   |

### DIFF
--- a/utils/index.js
+++ b/utils/index.js
@@ -153,7 +153,14 @@ const doYesNo = (claim, amount) => {
     }
 
     req.session.deductions[claim] = false
-    req.session.deductions[amount] = 0
+
+    if (amount && req.session.deductions[amount]) {
+      if (Object.keys(req.session.deductions[amount]).includes('amount')) {
+        req.session.deductions[amount].amount = 0.0
+      } else {
+        req.session.deductions[amount] = 0
+      }
+    }
 
     next()
   }
--- a/utils/index.js
+++ b/utils/index.js
@@ -127,6 +127,38 @@ const renderWithData = template => {
   }
 }
 
+/**
+ * Middleware to handle our yes/no question routing logic.
+ * If the yesNo page comes back "Yes"
+ * - set the session variable to "true"
+ * - redirect to the "/amount" url
+ *
+ * If the yesNo page comes back "No"
+ * - set the session variable to "false"
+ * - reset the "amount" var to 0
+ * - continue
+ *
+ * @param string claim the variable name with the claim
+ * @param string amount the variable name with the amount
+ */
+const doYesNo = (claim, amount) => {
+  return (req, res, next) => {
+    const claimVal = req.body[claim]
+
+    if (claimVal === 'Yes') {
+      req.session.deductions[claim] = true
+
+      // These two pages are hardcoded together
+      return res.redirect(`${req.path}/amount`)
+    }
+
+    req.session.deductions[claim] = false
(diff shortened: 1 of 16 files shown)


utils/index.js
```diff
@@ -157,6 +157,45 @@ const renderWithData = template => {
...
+ * If the yesNo page comes back "No"
+ * - set the session variable to "false"
+ * - reset the "amount" var to 0
+ * - continue
+ *
+ * @param string claim the variable name with the claim
+ * @param string amount the variable name with the amount
+ */
+const doYesNo = (claim, amount) => {
+  return (req, res, next) => {
```
katedee: lovely 👌 was just thinking about this, thanks for adding it

PR #143, opened 2019-09-17
Remove the old /auth routes + middleware

pcraig3:
It was a really super good idea and tons of excellent code, but unfortunately there is no room in the new reality for our tiered-authentication model.

Burn it all down. 🔥🔥🔥


pcraig3:
Worth pointing out that LGTM will love me for this.

PR #145, opened 2019-09-18
New back button layout

nmakuch:
### 1) Back button arrow has been switched to a left facing triangle `◀ Go Back`
| Before | After |
|--------|-------|
| Screen Shot 2019-09-18 at 18 51 02|    Screen Shot 2019-09-18 at 18 50 21   |

- Felt the arrow was taking up too much horizontal space
- The triangle is more visible and also follows [this govuk design pattern](https://design-system.service.gov.uk/components/back-link/)
- Default link underline felt really tight, so removed default `text-decoration` settings

### 2) Back button has been added to `base.pug` and moved above each pages main-heading

| Before | After |
|--------|-------|
|   Screen Shot 2019-09-18 at 18 55 28  |  Screen Shot 2019-09-18 at 18 55 55   |

- Since the back button is on every page it should probably be added to the layout component (not necessary, but cleaner)
- The back button was moved above the main page-headings as it is a separate element from the page content

### DIFF
--- a/views/_includes/backButton.pug
+++ b/views/_includes/backButton.pug
@@ -1,2 +1,2 @@
 if prevRoute
-  a(class='back-button' href=prevRoute.path) <span aria-hidden="true">← </span> Go back
+  a(class='back-button' href=prevRoute.path) <span aria-hidden="true">◀ </span> Go back
--- a/views/_includes/backButton.pug
+++ b/views/_includes/backButton.pug
@@ -1,2 +1,2 @@
 if prevRoute
-  a(class='back-button' href=prevRoute.path) <span aria-hidden="true">◀ </span> Go back
+  a(class='back-button' href=prevRoute.path) <span aria-hidden="true">◀ </span> #{__('Go back')}
--- a/views/base.pug
+++ b/views/base.pug
@@ -30,9 +30,12 @@ html(lang=getLocale() === 'fr' ? 'fr' : 'en')
       header
         include _includes/phaseBanner
         include _includes/fip
+      
       main
         if errors
           include _includes/errorList.pug
+
+        include _includes/backButton
 
         block content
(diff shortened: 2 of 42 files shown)


views/base.pug
```diff
@@ -30,10 +30,13 @@ html(lang=getLocale() === 'fr' ? 'fr' : 'en')
       header
         include _includes/phaseBanner
         include _includes/fip
+      
       main
         if errors
           include _includes/errorList.pug
 
+        include _includes/backButton
```
katedee: 👌

pcraig3: Cool, but let's put it above the errors.

views/_includes/backButton.pug
```diff
@@ -1,2 +1,2 @@
 if prevRoute
-  a(class='back-button' href=prevRoute.path) <span aria-hidden="true">← </span> Go back
\ No newline at end of file
+  a(class='back-button' href=prevRoute.path) <span aria-hidden="true">◀ </span> Go back
```
pcraig3: - the arrow could be applied with CSS (using `::before`) but since you have `aria-hidden="true"`; it's probably fine
- "Go back" should be translated. So like `#{__('Go back')}`. 

If you run the app in dev mode and then switch between english and french, it should should up in "/locales" for en.json and fr.json.

views/_includes/backButton.pug
```diff
@@ -1,2 +1,2 @@
 if prevRoute
```
pcraig3: Ah, is this why it doesn't show up sometimes?

Looks like 
- it's not on the start page (which we want)
- it's not on a page with an error (which we don't want ☹️)

See if you can figure out why it doesn't show up on the `/login/code` page when there's an error.

PR #146, opened 2019-09-20
Feature/check yr answers

katedee:
Putting this up for some initial thoughts/changes to be made.
There's still, unfortunately, quite a bit to do

There are a few ways to have this handled, but ultimately I went with having a separate array of objects of the lines that should be present on the page. We could handle it like income does, and go through the whole of user data and pick and choose from there, but that has it's own set of problems. I know this makes for another file to maintain though.

Initially, I had a crazy pug template that was handling a lot if/else for displaying the table row and formatting certain date. Now, I've offloaded most of that (except address) to:
- `checkAnswers` which is just a big array of objects, containing the relevant info to display + format. It's in it's own file just for a bit more ease in parsing the files
- `checkAnswersFormat`, which is where the bulk of our info formatting is happening, including whether a certain row should be passed along to our template.

**What this PR includes**
- displaying the Check Answers page, which included a bunch of data formatting, wew
- very basic cypress test that just goes to the page, hits the button
- more extensive cypress test that checks all the rows on the Check Answers page
- date-fns is added for date formatting
- some tests for the `formatAnswerInfo` function
-comparing the `urlPath` in checkAnswer to `routes.config` and make sure the urlPath we've put in actually exists

**What it won't include:**
- proper routing back to the Check Answers page. Honestly, the routing is it's own can of worms, and so I'll handle it in a separate PR

localhost_3005_checkAnswers (1)


routes/confirmation/checkAnswersFormat.js
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+  } else {
+    return (hasData(session,data[0],true) === null)
+      ? 'No'
+      : (hasData(session,data[0],true) === true)
+        ? 'Yes'
+        : (data[0].includes('Birth'))
+          ? dateFilter(hasData(session,data[0],true))
+          : hasData(session,data[0],true)
```
pcraig3: more

💡 it might be cleaner/clearer to do this as a `case` statement or whatever you call them in JavaScript.

routes/confirmation/confirmation.controller.js
```diff
@@ -1,10 +1,19 @@
...
   app.get('/review', renderWithData('confirmation/review'))
   app.post('/review', checkSchema(reviewSchema), checkErrors('confirmation/review'), doRedirect)
+
+  app.get('/checkAnswers', (req, res) => {
+    res.render('confirmation/check-answers', {
+      data: req.session,
+      prevRoute: getPreviousRoute(req),
+      answerInfo: formatAnswerInfo(req.session),
+    })
+  })
```
pcraig3: Maybe we could make `renderWithData` take an `options` object and then you could put them into the template.

routes/financial/financial.controller.js
```diff
@@ -22,5 +22,6 @@ const postConfirmIncome = (req, res, next) => {
     return res.redirect('/offramp/financial')
   }
 
+  req.session.incomeConfirmed = true
```
pcraig3: This is already set as `true` in the `api/user.json` file, so probably we should initially have it `null` or something.

utils/index.js
```diff
@@ -227,6 +227,35 @@ const hasData = (obj, key, returnVal = false) => {
...
+    day.length > 1 && 
+    day.charAt(0) == '0'
+  ) {
+    day = day.substr(1)
+  }
+  const monthIndex = Number(splitDate[1]) - 1
+  const year = splitDate[0]
+
+  return day + ' ' + monthNames[monthIndex] + ' ' + year
+}
```
pcraig3: I think we should start using a real date formatting method now.
We've used `date-fns` on other stuff, so that would be my preference: https://date-fns.org/v2.2.1/docs/format

(I have another date formatting func in here that I can refactor afterwards)

utils/index.js
```diff
@@ -262,7 +291,9 @@ const sortByLineNumber = (...objToSort) => {
  * @param {Array} routes array of route objects { name: "start", path: "/start" },
  * @returns { path: "" }
  */
-const getPreviousRoute = (path, session, routes = defaultRoutes) => {
+const getPreviousRoute = (req, routes = defaultRoutes) => {
+  const path = req.path
```
pcraig3: If you wanted to be really hip, you could be like
```js
const { path, session } = req
```

routes/confirmation/checkAnswers.js
```diff
@@ -0,0 +1,138 @@
+const answerInfo = [
+  { 
+    'sectionName': 'Personal Information',
+    'sectionLines': [
```
pcraig3: somewhat pedantic, but can we call this 'title' and 'row' or something a bit more terse?

routes/confirmation/checkAnswers.js
```diff
@@ -0,0 +1,138 @@
...
+      {
+        'text': 'Name',
+        'infoPath': ['personal.firstName', 'personal.lastName'],
+        'urlPath': '/personal/name',
+      },
+      {
+        'text': 'Date of Birth',
+        'infoPath': ['personal.dateOfBirth'],
+        'urlPath': '',
```
pcraig3: Interesting. We could also not have a url at all if we don't need it and then the template could check for it. 

But not super bothered either way -- this works too.

routes/confirmation/checkAnswers.js
```diff
@@ -0,0 +1,138 @@
...
+      {
+        'text': 'Public long-term care home',
+        'infoPath': ['deductions.trilliumLongTermCareClaim'],
+        'urlPath': '/trillium/longTermCare',
+      },
+      {
+        'text': 'Acommodation cost',
+        'infoPath': ['deductions.trilliumLongTermCareAmount'],
+        'urlPath': '/trillium/longTermCare/amount',
+        'displayIf': 'deductions.trilliumLongTermCareClaim',
```
pcraig3: Ooh, `displayIf` looks like a good way to handle this. 👍

views/confirmation/review.pug
```diff
@@ -75,6 +75,6 @@ block content
           input(id='review' name='review' type="checkbox" value="review" checked=(value =='review') aria-describedby=(errors ? `review-error` : false))
           label(for='review') #{__('I confirm that, as far as I know, all the information in this tax return is accurate')}
 
-    input#redirect(name='redirect', type='hidden', value='/confirmation')
+    input#redirect(name='redirect', type='hidden', value='/checkAnswers')
```
pcraig3: I think it's meant to be the other way around.

- `/deductions/climate-action-incentive`
- `/checkAnswers`
- `/review`
- `/confirmation`

public/scss/components/_breakdown-table.scss
```diff
@@ -94,4 +95,36 @@
...
+
+    @include md {
+      width: 35%;
+    }
+  }
+
+  .breakdown-table__row-link {
+    @include md {
+      width: 20%;
+    }
```
pcraig3: Let's bump the links down on mobile. 

Like this https://design-system.service.gov.uk/components/summary-list/default/index.html

pcraig3:
@katedee, also it says 'false' for home energy costs.  Probably not what we wanted.

image

routes/financial/financial.controller.js
katedee: oh woops, that's because of me. I should change it back 🤦‍♀

utils/index.js
katedee: oh, duh 🤦‍♀

routes/confirmation/checkAnswersFormat.js
```diff
@@ -0,0 +1,65 @@
...
+  } else {
+    if(hasData(session,data[0],true) === null || hasData(session,data[0],true) === false) {
+      return 'No'
+    } else if (hasData(session,data[0],true) === true) {
+      return 'Yes'
+    } else if (data[0].includes('Birth')) {
+      const initialDate = parseISO(hasData(session,data[0],true))
+      return format(new Date(initialDate), 'd MMMM yyyy')
+    } else {
+      return hasData(session,data[0],true)
```
pcraig3: Ah, sorry, I meant to do something like this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

I think it would be about as much work but cleaner — (but up to you).

routes/confirmation/checkAnswersFormat.js
```diff
@@ -0,0 +1,65 @@
...
+    const dataValues = []
+    data.map((val) => {
+      dataValues.push(hasData(session, val, true))
+    })
+
+    if (typeof dataValues[1] === 'number') {
+      return dataValues.reduce((a,b) => a + b, 0)
+    } else if (typeof dataValues[1] === 'string') {
+      return dataValues.join(' ')
+    }
```
pcraig3: Is it possible to not `return` here?

Ie, if `data.length > 1` is true, do we always have a number or a string?

katedee: so far we only have it for retrieving name (firstName, lastName), and then for addingNumbers.

I guess the better assumption would be splitting out into functions to handle that return val

routes/confirmation/checkAnswersFormat.js
```diff
@@ -0,0 +1,77 @@
...
+  data.map((val) => {
+    dataValues.push(hasData(session, val, true))
+  })
+
+  if (data.every( item => item.includes('Amount'))) {
+    return `$${currencyFilter(dataValues.reduce((a,b) => Number(a) + Number(b), 0))}`
+  } else if (data.every( item => typeof item === 'string')) {
+    return dataValues.join(' ')
+  } else {
+    throw new Error('Looks like you\'re trying to combine numebrs and strings  \n Are your infoPaths correct checkAnswers?')
```
pcraig3: looks like a typo — it should be "menbuns".

PR #147, opened 2019-09-23
task/Remove old HCL workflow file and write new ones in YML

pcraig3:
Migrating over from old HCL syntax to .yml syntax.

Should do the same things as before.

One thing to note is that I created two files: one for "test" and one for "test and deploy". I didn't see a clear way of doing this without creating two files, but if there is one, feel free to suggest it.

Also added a badge (👮) to the README! If the deploy fails, it should turn red. 🛑 

A good reference for this are the ones in here:

- https://github.com/cds-snc/cra-alpha/tree/master/.github/workflows


pcraig3:
Gonna merge since everything looks good over here.

PR #189, opened 2019-09-26
Fix/Allow single digit days and months for the date of birth field

pcraig3:
Small fix I noticed I was working on the security questions. 

Our validation is really picky right now: if you enter your date of birth as `9` instead of `09`, it will tell you you are wrong. This is a usability problem waiting to be discovered, so I figured I would preempt it. 

Allowed dob entries now:

- `09 09 1997` (same)
- `9 09 1997` (day has 1 digit)
- `09 9 1997` (month has one digit)
- `9 9 1997` (day and month have 1 digit)


schemas/login.schema.js
```diff
@@ -73,6 +73,12 @@ const isValidDay = {
   },
 }
 
+const toISOFormat = ({ dobYear, dobMonth, dobDay }) => {
+  const if0 = val => (val && val.length === 1 ? `0${val}` : val)
```
katedee: 👌

Issue #156, opened 2019-09-23
Please replace  https://claim-tax-benefits.azurewebsites.net/deductions/rrsp


YedidaZalik:
https://claim-tax-benefits.azurewebsites.net/deductions/rrsp
Please replace page with:
Heading: Deduct your Registered Retirement Savings Plan (RRSP) contributions

Did you contribute to a Registered Retirement Savings Plan (RRSP) before March 1, 2019?”
Yes
No
Continue  Cancel

nmakuch:
Closing as 190 got merged in

PR #190, opened 2019-09-26
Update /deductions/rrsp content

nmakuch:
### 1) Content update to the `/deductions/rrsp` page based on [Issue 156](https://github.com/cds-snc/cra-claim-tax-benefits/issues/156)

| Before | After |
|--------|-------|
| Screen Shot 2019-09-26 at 17 24 23  | Screen Shot 2019-09-26 at 17 24 30 |

- content was swapped in from [Issue 156](https://github.com/cds-snc/cra-claim-tax-benefits/issues/156)

### DIFF
--- a/public/scss/_forms.scss
+++ b/public/scss/_forms.scss
@@ -73,6 +73,12 @@ form.cra-form {
   input,
   select {
     @extend %form-element;
+    width: 300px;
+
+    @include sm {
+      width: 500px;
+    }
+
   }
 
   fieldset legend,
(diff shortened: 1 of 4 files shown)


public/scss/_forms.scss
```diff
@@ -73,6 +73,12 @@ form.cra-form {
   input,
   select {
     @extend %form-element;
+    width: 300px;
+
```
katedee: how did this sneak in here 🤔 
should we maybe consider a percentage value, and then set max-width in pixels? not a blocker, just something to think about

nmakuch: removed this stuff, I was trying to get the label larger because I didn't want it spilling on to 3 lines but it's really not that bad

nmakuch: also didn't realize it was making some inputs larger

PR #193, opened 2019-09-27
Feature/check answers back

katedee:
It's here!...kinda

I spent too much time overthinking this, and fighting this, and 118 lines definitely felt like a lifetime, but it's here. It mostly works for Yes/No, and will more than likely have to be re-evaluated for bigger edit pages (if we bring mailing address or maritalStatus back, etc)

### Basically:
- if we answer no, and it doesn't lead to an offramp, we go to checkAnswers
- if we answer Yes, and there's an amount page that follows, go to the amount page
- if we first go to the amount page, but hit the go back, I've made the assumption they want to review the yes/no to the deduction

This exists on the "go back" button and the submit button

checkAnswers


routes/deductions/deductions.spec.js
```diff
@@ -97,6 +97,7 @@ describe('Test /deductions responses', () => {
         url: '/trillium/studentResidence',
         key: 'trilliumStudentResidence',
         yesRedir: '/success',
+        noAmountPage: true,
```
pcraig3: This key is kind of the same as the other one. ie, if there is a "yesRedir", then there is no amount page.

Let's just use the existence of the other key to do this check.

routes/financial/financial.spec.js
```diff
@@ -29,4 +29,12 @@ describe('Test /financial responses', () => {
     expect(response.statusCode).toBe(302)
     expect(response.headers.location).toEqual('/deductions/rrsp')
   })
+
+  test('it redirects to the checkAnswers when posting Yes and having come from the checkAnswers page', async () => {
+    const response = await request(app)
+      .post('/financial/income')
+      .send({ confirmIncome: 'Yes', redirect: '/deductions/rrsp' }).query({ref: 'checkAnswers'})
```
pcraig3: Interesting, so I guess `.query` doesn't have to come directly after `.post`?

routes/personal/personal.spec.js
```diff
@@ -39,6 +39,14 @@ describe('Test /personal responses', () => {
       expect(response.headers.location).toEqual('/success')
       expect(response.statusCode).toBe(302)
     })
+
+    test('it redirects to the checkAnswers when posting Yes and having come from the checkAnswers page', async () => {
+      const response = await request(app)
+        .post('/personal/name').query({ref: 'checkAnswers'})
```
pcraig3: Let's either do 
-  `post().query().send()` or 
- `post().send().query()` 

just to be consistent.

utils/index.js
```diff
@@ -288,9 +297,38 @@ const getPreviousRoute = (req, routes = defaultRoutes) => {
     return oneRouteBack
   }
 
+  if(req.query && req.query.ref) {
+    //I'm assuming here we want to let them go back to yes/no
+    if( req.path.includes('amount') ) {
```
pcraig3: This works in some cases, not in others. I think it's a reasonable guess though.

So, if you do
- Did you pay rent? [Review]
- Yes (I want to go back to the "deduct your rent" page)
- Back

Then it makes sense. But if you do:

- Rent paid? [Review]
- Go back (indenting to go back to the check answers page)
- Did you pay rent?

Then maybe it's not what they expected. 

Anyway, it's kind of weird but it seems like it's reasonable. Normally, I would default to "don't add more if checks for edge cases" but it's okay as is.

utils/index.js
```diff
@@ -288,9 +297,38 @@ const getPreviousRoute = (req, routes = defaultRoutes) => {
...
+      return {path: `${prevRoute().path}?ref=checkAnswers`}
+    }
+    return routes.find(route => route.path === '/checkAnswers')
+  } 
+
   return prevRoute()
 }
 
+const returnToCheckAnswers = (req, res, claimYes = false, routes = defaultRoutes) => {
```
pcraig3: Do we need this param: `routes = defaultRoutes`? Looks like nobody is using it.

utils/index.js
```diff
@@ -288,9 +297,38 @@ const getPreviousRoute = (req, routes = defaultRoutes) => {
...
+    'editInfo' !== 'skip' &&
+    claimYes
+    ) {
+      return res.redirect(`${nextRoute.path}?ref=checkAnswers`)
+    } else if (
+      ref === 'checkAnswers' && 
+      !claimYes 
+    ) {
+      return res.redirect(`/checkAnswers`)
+    }
```
pcraig3: This logic is a bit weird because it doesn't cover all cases.

Really we want to say:

- if !checkAnswers (what do we return? error?)
- if editInfo + !skip + claimYes (return path?ref=checkAnswers)
- else (redirect)

utils/index.js
```diff
@@ -288,9 +297,38 @@ const getPreviousRoute = (req, routes = defaultRoutes) => {
...
 }
 
+const returnToCheckAnswers = (req, res, claimYes = false, routes = defaultRoutes) => {
+
+  const currentRoute = getRouteWithIndexByPath(req.path)
+  const nextRoute = routes[currentRoute.index + 1]
+  const ref = req.query.ref
+  
+  if(
+    ref === 'checkAnswers' &&
```
pcraig3: Okay, so ref can only every be `checkAnswers`, right?

utils/index.spec.js
```diff
@@ -7,6 +7,7 @@ const testRoutes = [
   { name: 'rrsp', path: '/deductions/rrsp' },
   { name: 'rrsp amount', path: '/deductions/rrsp/amount', editInfo: 'deductions.rrspClaim' },
   { name: 'medical', path: '/deductions/medical' },
```
pcraig3: I think you can remove the `name:` keys here?

utils/index.js
katedee: I can get rid of it, no problem

utils/index.js
katedee: `const nextRoute = routes[currentRoute.index + 1]`

utils/index.js
katedee: so we're not really hitting this _unless_ there's a ref of checkAnswers, but I'll add to make that explicit where it could be called (ex in doRedirect and doYesNo)

routes/financial/financial.spec.js
katedee: oops, good catch!

utils/index.js
katedee: we're using it for checking the nextRoute

pcraig3: Yeah, but we don't need it to be a param though, right? On line 316, you could just say `defaultRoutes[currentRoute.index + 1]`

utils/index.js
pcraig3: Since we're checking externally now, let's make the checks here less strict. (ie, we don't need to check that ref === `checkAnswers`.)

utils/index.js
```diff
@@ -310,21 +310,19 @@ const getPreviousRoute = (req, routes = defaultRoutes) => {
...
   if(
-    ref === 'checkAnswers' && 
     'editInfo' in nextRoute &&
     'editInfo' !== 'skip' &&
     claimYes
     ) {
       return res.redirect(`${nextRoute.path}?ref=checkAnswers`)
     } else if (
-      ref === 'checkAnswers' && 
       !claimYes
```
pcraig3: This is now now true by virtue of the first check (ie, it has to be truthy for the initial check to pass)

pcraig3: If we wanted to have a case where 
('editInfo' in nextRoute ||  'editInfo' === 'skip') was NOT true and we would do something else, then that second check makes sense (!claimYes), but not currently.

PR #194, opened 2019-09-27
Fix/values unsaved

katedee:
more accurately, it was undisplayed values, as they _are_ being saved to session

- save the yes/no values on personal pages (ex. marital, address)
- display those saved values in the yes/no buttons. We were checking for value === 'Yes' or value === 'No', but we're storing as true/false

confirmsave


routes/deductions/deductions.controller.js
```diff
@@ -29,6 +29,10 @@ module.exports = function(app) {
     checkSchema(rrspSchema),
     checkErrors('deductions/rrsp'),
     doYesNo('rrspClaim', 'rrspAmount'),
```
pcraig3: So, maybe this should be clearer but the second parameter in `doYesNo` actually clears the value if it's there. (because of the repetitive nature of this stuff).

https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/utils/index.js#L158

So we don't need to so this here or the others.

routes/deductions/deductions.controller.js
```diff
@@ -73,6 +81,10 @@ module.exports = function(app) {
     checkSchema(medicalExpenseSchema),
     checkErrors('deductions/medical'),
     doYesNo('medicalExpenseClaim', 'medicalExpenseAmount'),
```
pcraig3: I think if you changed this to "medicalExpenseClaimAmount", it should work out.

pcraig3: But the actual pug template might be using a different name than the user.json value.

Issue #166, opened 2019-09-23
** Note : this is for two different pages, page:/  https://claim-tax-benefits.azurewebsites.net/trillium/rent and page:/ https://claim-tax-benefits.azurewebsites.net/trillium/rent/**amount


YedidaZalik:
** Note : this is for two different pages, page:/ https://claim-tax-benefits.azurewebsites.net/trillium/rent and page:/ https://claim-tax-benefits.azurewebsites.net/trillium/rent/**amount

Please  replace with several new pages:

New Page 
Heading:  The Ontario Trillium Benefit
The Ontario Trillium Benefit (OTB) is a government payment that helps with energy, sales and property taxes. You can learn about the OTB at https://www.canada.ca/en/revenue-agency/services/child-family-benefits/provincial-territorial-programs/ontario-trillium-benefit-questions-answers.html

Apply for OTB on the pages that follow.

Continue Cancel

New Page

Heading: Apply for OTB:rent payments

Did you pay rent for the place you lived most of 2018?

Yes 
No 

Continue Cancel

New Page
Heading: 
Enter your rent payments to apply for OTB

The government might ask for your rent receipts.

How much total rent did you pay for the place you lived most of 2018?

Continue Cancel

nmakuch:
closing as 199 has been merged

PR #199, opened 2019-09-27
Update to the trillium section

nmakuch:
### `/trillium/start` page

Screen Shot 2019-09-27 at 18 36 57

### `/trillium/rent` page

| Before | After |
|--------|-------|
|  Screen Shot 2019-09-27 at 18 32 30 | Screen Shot 2019-09-27 at 18 32 37 |

### `/trillium/rent/amount` page

| Before | After |
|--------|-------|
| Screen Shot 2019-09-27 at 18 33 30| Screen Shot 2019-09-27 at 18 33 37  |

- content was swapped in from [Issue 166](https://github.com/cds-snc/cra-claim-tax-benefits/issues/166)

### DIFF
--- a/cypress/integration/full_run_simple.spec.js
+++ b/cypress/integration/full_run_simple.spec.js
@@ -188,7 +188,7 @@ describe('Full run through', function() {
   })
 
   it('navigates the Trillium Start page', function() {
-    //TRILLIUM RENT
+    //TRILLIUM START PAGE
     cy.url().should('contain', '/trillium/start')
     cy.get('h1').should('contain', 'The Ontario Trillium Benefit')
 
--- a/cypress/integration/full_run_simple.spec.js
+++ b/cypress/integration/full_run_simple.spec.js
@@ -187,10 +187,20 @@ describe('Full run through', function() {
       .click()
   })
 
+  it('navigates the Trillium Start page', function() {
+    //TRILLIUM RENT
+    cy.url().should('contain', '/trillium/start')
+    cy.get('h1').should('contain', 'The Ontario Trillium Benefit')
+
+    cy.get(`a[href="/trillium/rent"]`)
+      .should('contain', 'Continue')
+      .click()
+  })
+
   it('navigates the Trillium Rent page', function() {
     //TRILLIUM RENT
     cy.url().should('contain', '/trillium/rent')
-    cy.get('h1').should('contain', 'Deduct your rent payments')
+    cy.get('h1').should('contain', 'Apply for OTB: rent payments')
 
     cy.get('input#trilliumRentClaimNo + label').should('have.attr', 'for', 'trilliumRentClaimNo')
 
--- a/views/_includes/buttonMixin.pug
+++ b/views/_includes/buttonMixin.pug
@@ -5,9 +5,12 @@ mixin formButtons(submitButtonText = 'Continue', cancelUrl='/clear')
     .buttons--right
       a.transparent.full-width(href=cancelUrl role='button' draggable='false') #{__('Cancel')}
 
-mixin linkButtons(confirmURL)
+mixin linkButtons(confirmURL, continueButton)
   .buttons-row
-    a(href=confirmURL role='button' draggable='false') #{__('Confirm')}
+    if continueButton
+      a(href=confirmURL role='button' draggable='false') #{__('Continue')}
+    else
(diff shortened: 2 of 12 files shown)


views/_includes/buttonMixin.pug
```diff
@@ -5,9 +5,12 @@ mixin formButtons(submitButtonText = 'Continue', cancelUrl='/clear')
...
       a.transparent.full-width(href=cancelUrl role='button' draggable='false') #{__('Cancel')}
 
-mixin linkButtons(confirmURL)
+mixin linkButtons(confirmURL, continueButton)
   .buttons-row
-    a(href=confirmURL role='button' draggable='false') #{__('Confirm')}
+    if continueButton
+      a(href=confirmURL role='button' draggable='false') #{__('Continue')}
+    else
+      a(href=confirmURL role='button' draggable='false') #{__('Confirm')}
```
pcraig3: The only difference between these two seems to be the button text. 

Let's just pass in the button text, but default to 'Confirm', like we do here: https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/views/_includes/buttonMixin.pug#L1

views/deductions/trillium-start.pug
```diff
@@ -0,0 +1,16 @@
...
+block variables
+  -var title = __('The Ontario Trillium Benefit')
+  -var trilliumRentClaim = !hasData(data, 'deductions.trilliumRentClaim') ? '' : data.deductions.trilliumRentClaim ? 'Yes' : 'No'
+
+block content
+
+  h1 #{title}
+
+  div
+    p #{__('The Ontario Trillium Benefit (OTB) is a government payment that helps with energy, sales and property taxes. You can learn about the OTB at')} <a href='https://www.canada.ca/en/revenue-agency/services/child-family-benefits/provincial-territorial-programs/ontario-trillium-benefit-questions-answers.html'>https://www.canada.ca/en/revenue-agency/services/child-family-benefits/provincial-territorial-programs/ontario-trillium-benefit-questions-answers.html</a>
```
pcraig3: the way we do links is that we want to put the actual `href` in as a translateable string as well, like we do here: https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/views/_includes/offramp-bullets.pug#L2

That way, when we switch the page to french, the url also changes.

views/deductions/trillium-start.pug
```diff
@@ -0,0 +1,16 @@
...
+
+    p #{__('Apply for OTB on the pages that follow.')}
```
pcraig3: Are we sure it's not `Apply for the OTB`?

views/deductions/trillium-start.pug
```diff
@@ -0,0 +1,16 @@
+extends ../base
+
+block variables
+  -var title = __('The Ontario Trillium Benefit')
+  -var trilliumRentClaim = !hasData(data, 'deductions.trilliumRentClaim') ? '' : data.deductions.trilliumRentClaim ? 'Yes' : 'No'
```
pcraig3: We can get rid of this.

views/deductions/trillium-rent-amount.pug
```diff
@@ -1,7 +1,7 @@
 extends ../base
 
 block variables
-  -var title = __('Enter rent paid in 2018')
+  -var title = __('Enter your rent payments to apply for OTB')
```
pcraig3: I'm not sure if they are actually "applying". I thought they would just be automatically eligible and we are asking for their rent payments to see how much they get back.

Can you check with @YedidaZalik?

YedidaZalik: I used "apply" because of #4 in this: https://www.canada.ca/en/revenue-agency/services/child-family-benefits/provincial-territorial-programs/ontario-trillium-benefit-questions-answers.html where it says "To apply for the 2019 OEPTC and/or NOEC components on your 2018 return, fill out  . . ."  Is there a way to flag that for Nancy to double-check?

cypress/integration/full_run_simple.spec.js
```diff
@@ -187,10 +187,20 @@ describe('Full run through', function() {
       .click()
   })
 
+  it('navigates the Trillium Start page', function() {
+    //TRILLIUM RENT
```
pcraig3: This comment is wrong 🙅‍♂

PR #201, opened 2019-09-30
Feature/Security question page & 2 questions

pcraig3:
**Note: the content isn't final — I just didn't want to have this sit any longer.**

---

This PR is for adding a screen with security questions, and then the questions themselves.
It was taking a while, so I only added 2 questions.

Lots of stuff happening:
- we're adding new pages outside of a linear flow
- the form pages are more complex than normal
- a bunch of content so had to re-write some existing tests
- found some tiny validation bugs 
- refactored the DoB CSS to make it more reusable

Look through and see if it functions properly. 
Should be accessible and have validation messages where needed.
The words we can change later (tomorrow), but for now let's get it in if it's working.

Unlike the previous pages in the login flow, we're just that the values make sense (not empty, dates aren't '0', etc) but we're not matching it against anything or saving the values. The only thing we remember is the url of the question that they picked. The stuff you enter to get past that screen is just lost. (and it's supposed to work that way.)

## gif

This is all new, so there's no "before", only an "after".

q


pcraig3:
> **new alerts:**
> 
> * 1 for Server-side URL redirect

This isn't really a user-supplied value. Our URLs are validated against a whitelist. Might change the code to reflect this, but don't need to right now.

PR #215, opened 2019-10-04
Add body trimmer to all form requests

pcraig3:
We pretty much never want whitespace in form requests, so we should trim all inputs by default.

Heard about this happening in usability testing so let's make this change.

Previously, this would fail:

- day: `' 9 '`
- month: `' 9 '`
- year: `' 1977 '`

Now it works.

Added a unit test to confirm behaviour.

### DIFF
--- a/app.js
+++ b/app.js
@@ -5,6 +5,7 @@ const globalError = require('http-errors')
 // import node modules.
 const express = require('express'),
   cookieParser = require('cookie-parser'),
+  trimRequest = require('trim-request'),
   compression = require('compression'),
   helmet = require('helmet'),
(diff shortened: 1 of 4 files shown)


app.js
```diff
@@ -71,6 +72,7 @@ app.use(helmet.contentSecurityPolicy({ directives: csp }))
 
 // gzip response body compression.
 app.use(compression())
+app.use(trimRequest.all)
```
katedee: ✨

pcraig3:
Killed the Dockerfile job because it was hanging. No changes to the Dockerfile in this PR.

PR #226, opened 2019-10-08
removed ide-specific configuration file from version control

charlesmorin:
This PR simply removes the .vscode folder from version control. Git repositories should normally remain IDE agnostic, especially from a CI/CD perspective, which isn't going to use an IDE to build the app.


pcraig3:
Approved!

PR #230, opened 2019-10-10
Fix/Whitelist Redirects

pcraig3:
Whitelist redirect URLs in `doRedirect` middleware. 

@obrien-j's LGTM thing was warning us that redirecting to whatever was posted in the form was not super safe. I get it, fair enough. 

So, since we now have a list of all the paths we use in the application, let's match the redirect to a specific path. 

The first commit is the actual change. The second commit is updating the 90 unit tests or so that were broken. (Lots of our redirect tests were using `/` or `/success`, so I've mostly changed them to `/start`.)

@katedee, you're probably the best one to review this.

### DIFF
--- a/utils/index.js
+++ b/utils/index.js
@@ -109,7 +109,7 @@ const checkErrors = template => {
 // POST functions that handle setting the login data in the session and will redirecting to the next page or send back an error to the client.
 // Note that this is not the only error validation, see routes defined above.
 const doRedirect = (req, res) => {
-  let redirect = req.body.redirect || null
+  const redirect = req.body.redirect || null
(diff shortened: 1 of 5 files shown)


utils/index.js
```diff
@@ -109,7 +109,7 @@ const checkErrors = template => {
 // POST functions that handle setting the login data in the session and will redirecting to the next page or send back an error to the client.
 // Note that this is not the only error validation, see routes defined above.
 const doRedirect = (req, res) => {
-  let redirect = req.body.redirect || null
+  const redirect = req.body.redirect || null
```
katedee: 👌

PR #232, opened 2019-10-10
Feature/dob mixin

katedee:
Debated a few ways of doing this (container mixin, with then mixins for the individual date inputs, so we could leverage `&attributes`), but ended up settling on this way of handling it, which allows for passing in the fewest things in the mixin. Might want to add something to make autocomplete option though.

- `idPrefix` - since each input will end with day/month/year, we're just taking and (ideally short) id prefix to append

- `legendText` - the text for the legend

- `isoHintText` - the iso value for the hint text that comes along with the date fields

- `values` - I decided to let passing in the values in case we have multiple date fields on the page (as I believe we will). However, right now it's such that it assumes it will be passed in as day, month, year. Open to changing this and running a check to make sure we match correctly, but I also think we should keep date fields order consistent in the app. Though I'm curious as to what others think!

## Edits
- change `isValidDay` to just check for an int between 1-31, since in all our cases we're comparing against an existing date, we don't need to go wild here
- only put a red border on the field that has the error, or on all of them if the date does not match
- make sure that the inline error list, only lists the associated date errors (and not any others present on the entire form)

Screen Shot 2019-10-10 at 3 37 12 PM

---

image

---

image

### DIFF
--- a/schemas/login.schema.js
+++ b/schemas/login.schema.js
@@ -58,24 +58,13 @@ const lastDayInMonth = (year, month) => {
   return new Date(year, month + 1, 0).getDate()
 }
 
-const isValidDay = {
-  errorMessage: 'errors.login.dateOfBirth.validDay',
-  validate: (value, req) => {
-    const yearKey = Object.keys(req.body).filter((key) => /Year/.test(key))
-    const monthKey = Object.keys(req.body).filter((key) => /Month/.test(key))
-
-    const year = parseInt(req.body[yearKey], 10)
-    let month = parseInt(req.body[monthKey], 10)
-    const day = parseInt(value, 10)
-
-    if (!day || !month || !year) {
-      return false
-    }
-
-    //subtract one because Date for months starts at a 0 index for Jan 🤓
-    month -= 1
-    return day >= 1 && day <= lastDayInMonth(year, month)
-  },
+const isValidDay = (errorMessageString = 'errors.login.dateOfBirth.validDay') => {
+  return {
+    isInt: {
+      errorMessage: errorMessageString,
+      options: { min: 1, max: 31 },
+    },
+  }
 }
 
 const toISOFormat = ({ dobYear, dobMonth, dobDay }) => {
@@ -98,7 +87,8 @@ const isMatchingDoB = {
 
 const dobSchema = {
   dobDay: {
-    ...validationArray([isValidDay, isMatchingDoB]),
+    ...isValidDay(),
+    ...validationArray([isMatchingDoB]),
   },
   dobMonth: {
     isInt: {
@@ -130,9 +120,7 @@ const childSchema = {
       negated: true,
     },
   },
-  dobDay: {
-    ...validationArray([isValidDay]),
-  },
+  dobDay: isValidDay(),
   dobMonth: {
     isInt: {
       errorMessage: 'errors.login.dateOfBirth.validMonth',
@@ -148,9 +136,7 @@ const childSchema = {
 }
 
 const dateOfResidenceSchema = {
-  dobDay: {
-    ...validationArray([isValidDay]),
-  },
+  dobDay: isValidDay(),
   dobMonth: {
     isInt: {
       errorMessage: 'errors.login.dateOfBirth.validMonth',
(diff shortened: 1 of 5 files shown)


views/_includes/input-dates.pug
```diff
@@ -0,0 +1,24 @@
...
+      legend
+        h2 #{__(legendText)}
+
+      span#dobHint.cra-form-message #{__('For Example:')} #{isoDateHintText(isoHintText)}
+
+      if errors
+        ul
+          each err in errors
+            li
+              +validationMessage(err.msg, err.param)
```
pcraig3: This starts working as soon as there are errors, but we have some pages where the date part is only part of the errors that are possible.

pcraig3: https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/views/login/questions/child.pug#L39

views/_includes/input-dates.pug
```diff
@@ -0,0 +1,24 @@
...
+      div.w-1-4.d--ib
+        label(for=`${idPrefix}Day`) #{__('Day')}
```
pcraig3: "Prefix" is a nice idea, but I abandoned it previously because our "matching" code in the validation only works with the dobStuff: https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/schemas/login.schema.js#L64-L65

Potential solutions: 
- Update the validations so that we can pass through prefixes somehow
- Keep everything as some generic looking `day`, `month`, `year` type thing
- Keep everything as `dobDay` all the time.

schemas/login.schema.js
```diff
@@ -54,25 +54,13 @@ const sinSchema = {
...
-    month -= 1
-    return day >= 1 && day <= lastDayInMonth(year, month)
-  },
+const isValidDay = (errorMessageString = 'errors.login.dateOfBirth.validDay') => {
+  return {
+    isInt: {
+      errorMessage: errorMessageString,
+      options: { min: 1, max: 31 },
+    },
+  }
```
pcraig3: 😚👌

I am thrilled about this.

views/_includes/input-dates.pug
```diff
@@ -0,0 +1,28 @@
+mixin dateInputs(idPrefix, legendText, isoHintText, values)
```
pcraig3: What happens if there is no `idPrefix`? Can we set a default value?

views/_includes/input-dates.pug
pcraig3: Looks like removing the validation for leap years solves this: 👍

views/_includes/input-dates.pug
katedee: ah, good catch

katedee: oh wait, this actually won't be an issue because you have to pass something or a null to continue

views/_includes/input-dates.pug
```diff
@@ -0,0 +1,28 @@
...
+            if err.param.includes(idPrefix)
+              li
+                +validationMessage(err.msg, err.param)
+      div.w-1-4.d--ib(class={['has-error--inline']: errors && errors[`${idPrefix}Day`]})
+        label(for=`${idPrefix}Day`) #{__('Day')}
+        input#dobDay(name=`${idPrefix}Day`, autofocus, type='text', value=values[0], autocomplete=`${idPrefix}-day`, aria-describedby=(errors && errors[`${idPrefix}Day`] ? `${idPrefix}Day-error` : false))
+
+      div.w-1-4.d--ib(class={['has-error--inline']: errors && errors[`${idPrefix}Month`]})
+        label(for=`${idPrefix}Month`) #{__('Month')}
+        input#dobMonth(name=`${idPrefix}Month`, type='text', value=values[1], autocomplete=`${idPrefix}-month`, aria-describedby=(errors && errors[`${idPrefix}Month`] ? `${idPrefix}Month-error` : false))
```
pcraig3: This autocomplete one is a bit weird: 
https://design-system.service.gov.uk/components/date-input/#use-the-autocomplete-attribute-for-a-date-of-birth

Basically, if we're asking for a date of birth for the current user, we should include this, and then hardcode the values (as `bday-day`, `bday-month`, `bday-year`). Otherwise, we shouldn't show the autcomplete at all. We could use the `dob` idPrefix to decide, or another var (defaults to false).

We could also drop support for this and circle back later.

PR #235, opened 2019-10-10
Add Azure Application Insights for telemetry purposes

charlesmorin:
### This PR consists in
- adding Azure App Insights dependency to gather metrics and tracing logs.
- register the App Insights through an instrumentation key that is defined in Azure and injected as an environment variable only when live (will not register when deploying locally).
- added documentation as for deploying to a new custom Azure App Service for testing purposes.

### DIFF
--- a/DEPLOY.md
+++ b/DEPLOY.md
@@ -127,7 +127,7 @@ You can then find the URL in the JSON response or in the Azure portal. URL shoul
 
 When finished testing, you can delete the Azure App Service webapp using the following command:
 
-`az webapp delete --resource-group cracdscollaborationrg --name claim-tax-benefits-{tag_name}`
+`az webapp delete --resource-group cdscracollab-innovation-rg --name claim-tax-benefits-{tag_name}`
 
 ### All done!
 
--- a/DEPLOY.md
+++ b/DEPLOY.md
@@ -42,7 +42,7 @@ Make sure you have the [docker command line interface](https://docs.docker.com/e
 First, run `docker build` and pass in an optional `GITHUB_SHA_ARG`.
 
 ```
-build -t base --build-arg GITHUB_SHA_ARG=codename_cobra .
+docker build -t base --build-arg GITHUB_SHA_ARG=codename_cobra .
 ```
 
 #### Optional env var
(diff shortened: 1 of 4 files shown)


DEPLOY.md
```diff
@@ -42,7 +42,7 @@ Make sure you have the [docker command line interface](https://docs.docker.com/e
 First, run `docker build` and pass in an optional `GITHUB_SHA_ARG`.
 
 ```
-build -t base --build-arg GITHUB_SHA_ARG=codename_cobra .
+docker build -t base --build-arg GITHUB_SHA_ARG=codename_cobra .
```
pcraig3: Good catch.

DEPLOY.md
```diff
@@ -117,8 +117,20 @@ You can also update the app by using the user interface. For a one-off update, t
...
+
+In case you don't want to override the live version of the app, you can spin a new instance of App Service for testing purposes by using the following command:
+
+`az webapp create --resource-group cdscracollab-innovation-rg --plan alphaPlan --name claim-tax-benefits-{tag_name} --deployment-container-image-name cdssnc/cra-claim-tax-benefits:{tag_name}`
+
+You can then find the URL in the JSON response or in the Azure portal. URL should nornmally be `https://claim-tax-benefits-{tag_name}.azurewebsites.net`
+
+When finished testing, you can delete the Azure App Service webapp using the following command:
+
+`az webapp delete --resource-group cracdscollaborationrg --name claim-tax-benefits-{tag_name}`
```
pcraig3: I think the resource group has the wrong name here.

app.js
```diff
@@ -1,9 +1,11 @@
+
```
pcraig3: Let's remove this empty line

app.js
```diff
@@ -28,13 +30,18 @@ const express = require('express'),
 // initialize application.
 var app = express()
 
+if (process.env.NODE_ENV !== 'test') {
+  // register to Azure Application Insights service for telemetry purposes
+  // instrumention key is provisioned from Azure App Service application setting (env variable)
+  azureApplicationInsights.setup(process.env.APPINSIGHTS_INSTRUMENTATIONKEY).start()
```
pcraig3: I think we should only add this if the `process.env.APPINSIGHTS_INSTRUMENTATIONKEY` env var exists.
That way, it will still build on Heroku and locally without having this in the `.env`.

Also, I don't want to be too prescriptive, but I think it's better to run the Azure Analtyics stuff only in 'prod' mode.

So the code would be:

```
if (process.env.NODE_ENV === 'production' && process.env.APPINSIGHTS_INSTRUMENTATIONKEY) {
// init app insights
}
```

The morgan logger we still want to run in dev mode, so we can leave that how it was before.

DEPLOY.md
charlesmorin: good catch, fixing it

app.js
charlesmorin: sounds good!

charlesmorin: it's all good. Done.

PR #251, opened 2019-10-11
minor change to log format for dev and production runtimes

charlesmorin:
This PR consists in using the `dev` log format in morgan middleware when in dev mode, and `combined` log format when deployed to production.


charlesmorin:
Merging as LGTM seems to be down or something. Tested and working.

PR #271, opened 2019-10-17
added Terraform infrastructure as code script

charlesmorin:
### This PR consists in

- adding a /scripts folder into which we will store Infrastructure as Code scripts moving forward.
- adding a new script to create an Azure AppService along with an Azure Application Insights resource to easily deploy a new version of the Claim Tax Benefits within a Continuous Delivery (CD) pipeline.

Usage documentation can be found in `USAGE.md`

**No impact on application's code.**


scripts/appservice/create-appservice.tfvars
```diff
@@ -0,0 +1,7 @@
+location = "canada central"
+resource_group_name = "cdscracollab-innovation-rg"
+appservice_plan_id = "alphaPlan"
+appservice_name = "claim-tax-benefits-test"
+docker_image_name = "cdssnc/cra-clai m-tax-benefits:latest"
```
pcraig3: Extra space here that we don't want.

scripts/appservice/create-appservice.tf
```diff
@@ -0,0 +1,75 @@
...
+  application_type    = "web"
+}
+
+# Create Azure AppService
+resource "azurerm_app_service" "main" {
+  name                = "${var.appservice_name}"
+  location            = "${var.location}"
+  resource_group_name = "${var.resource_group_name}"
+  app_service_plan_id = "${var.appservice_plan_id}"
+
```
pcraig3: Let's turn on `https_only` by default: https://www.terraform.io/docs/providers/azurerm/r/app_service.html#https_only

scripts/appservice/create-appservice.tfvars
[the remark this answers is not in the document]
charlesmorin: good catch!

charlesmorin:
- folder renamed to `/scripts/terraform`
- added `.gitignore` in `/scripts/terraform`. Any reason why we keep it in a separate `gitignore` instead of expanding the root `gitignore`? I don't mind to be honest but what just curious.
- good point for the `tfvars` file name. Didn't catch that in the doc! Awesome! We might need to segragate that into subfolders as we put in more terraform scripts since the `tfvars` file might need to be different.
- As for the plan and resource group, I would agree if we were granted permission to create RGs and Plans in Azure. These are provided by the CRA Cloud CoE. Given that context, I believe we have two choices: keep it the current way (assuming they are provided from another team), or enhance the script to check if the plan and RG exist, and in the negative, have it to create those resources for us. I know for sure than plan and RGs will be handled by Cloud CoE in the future, so I would suggest status quo on that one.

pcraig3:
Not sure why the tests aren't running. Feel free to merge.

PR #292, opened 2019-10-21
Fix/Replace cookie session with memorystore

pcraig3:
Okay, so on Friday I was looking to save the response ID, and it opened up this whole can of worms. 

I tried saving it as a cookie, which worked in a browser window, but not as an iframe. So that was out. Then I tried saving it to the session and found out that saving things to the session wasn't working, which is the opposite of [what this trello card actually says](https://trello.com/c/pgml86gO/268-l-%F0%9F%95%B7bug-actually-delete-your-info).

> Deleting the session server-side doesn't actually delete your saved info. For some reason, the cookie stays around forever.

Anyway, the long&short is that sessions weren't being saved at all because the session cookie wasn't being set, so trying to save a response ID to the session wasn't working. Over time our session data has grown, and [became large enough that it was too big to save as a cookie](https://github.com/cds-snc/cra-claim-tax-benefits/pull/292/commits/04985b2f481187d9a7275aff8e7758dc1049daae).

Switching from `cookie-session` to `memorystore` fixes this and kills the bug.


utils/index.js
```diff
@@ -61,7 +61,10 @@ const checkPublic = function(req, res, next) {
   // check if user exists in session (ie, by checking for firstName)
   const { personal: { firstName = null } = {} } = req.session
   if (!firstName) {
-    req.session = API.getUser('A5G98S4K1')
+    const user = API.getUser('A5G98S4K1')
+
+    // setting req.session = {obj} causes an error, so assign the keys one at a time
+    Object.keys(user).map(key => (req.session[key] = user[key]))
```
katedee: I can't believe I didn't think of this when I was working on memorystore 🤦‍♀

pcraig3: Yeah, I was briefly considering changing all data to a new key (`session.user`, maybe?).

Anyway, that would have been awful.

PR #302, opened 2019-10-22
fixed SonarQube bugs and code smells, and added back a missing Cypress script

charlesmorin:
**This PR consists in**
- fixing 3 code smells
- fixing 2 bugs
- adding a script that was lost in previous merges

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -4,7 +4,7 @@ var API = (function(userFound) {
   const _user = userFound
 
   const getUser = code => {
-    if (code === userFound.login.code) {
+    if (code === _user.login.code) {
       return _user
     }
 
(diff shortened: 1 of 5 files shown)


api/index.js
```diff
@@ -1,10 +1,10 @@
 const user = require('./user.json')
 
-var API = (function(user) {
-  const _user = user
+var API = (function(userFound) {
+  const _user = userFound
 
   const getUser = code => {
-    if (code === user.login.code) {
+    if (code === userFound.login.code) {
```
pcraig3: Let's make this `code === _user.login.code`. I think this was an oversight.

package.json
```diff
@@ -36,6 +36,7 @@
     "cypress": "start-server-and-test start http://localhost:3005 cypress:open",
```
pcraig3: Let's have this one use `start:cypress` as well

PR #310, opened 2019-10-23
🌈new and improved prevroute

katedee:
✨ it  can skip multiple pages at a time
 ✨ we no longer have to separately return based on index if it's one route back or five routes back

since we'll be breaking up some pages into multiple page flows in the future, there's a potential need to skip multiple pages beyond just the survey test

### DIFF
--- a/utils/index.js
+++ b/utils/index.js
@@ -285,9 +285,18 @@ const getPreviousRoute = (req, routes = defaultRoutes) => {
   }
 
   const prevRoute = () => {
+    /**
+     * filter to the only possible previous routes (so all routes before this one)
+     * reverse the order so when we look over them, we go "backward"
+    */
     const allPreviousRoutes = routes.slice(0, route.index).reverse()
     
     const previousRoute = allPreviousRoutes.find((route) => {
+       /**
+       * essentially check if the page before
+       * - is an edit page, that has been edited
+       * - or is not an edit page, and is the next logical page
+      */
       if(
       'editInfo' in route &&
       hasData(session, route.editInfo, true) ||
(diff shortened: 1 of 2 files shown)


utils/index.js
```diff
@@ -285,23 +285,19 @@ const getPreviousRoute = (req, routes = defaultRoutes) => {
...
-      'editInfo' in oneRouteBack &&
-      !hasData(session, oneRouteBack.editInfo, true)
-    ) {
-      return routes[Number(route.index) - 2] || false
-    }
-
-    return oneRouteBack
+    const allPreviousRoutes = routes.slice(0, route.index).reverse()
+    
+    const previousRoute = allPreviousRoutes.find((route) => {
```
pcraig3: Can we leave the comments in saying what we're checking for?

katedee: sure!

utils/index.js
```diff
@@ -285,23 +285,28 @@ const getPreviousRoute = (req, routes = defaultRoutes) => {
...
+       /**
+       * essentially check if the page before
+       * - is an edit page, that has been edited
+       * - or is not an edit page, and is the next logical page
+      */
+      if(
+      'editInfo' in route &&
+      hasData(session, route.editInfo, true) ||
+      !route.hasOwnProperty('editInfo')
+      ) {
```
pcraig3: Oops, can we add brackets here to make sure it works how we want.

Like

```
if(
      ('editInfo' in route && hasData(session, route.editInfo, true)) || !route.hasOwnProperty('editInfo')
      )
```

PR #312, opened 2019-10-23
Offramp survey banners + New Page at end of flow

nmakuch:
### Created the offramp-banner _include
- Showing up across multiple pages so made this into an include
- Spans bottom of page under footer
- Link leads to new page

### Created an /offramp/confirmation page that tells users the prototype has ended
- Unstyled
- Instructions to click outside of window

### Added medical expense ($640) to the Yes/No label
- Used to be 'X'

Screen Shot 2019-10-24 at 15 13 10

### GIF

offramp


views/offramp/confirmation.pug
```diff
@@ -0,0 +1,16 @@
...
+
+block content
+
+  h1 #{title}
+ 
+  div
+    p #{__('To continue the survey click anywhere outside this popup window')}
+    p #{__('Clicking')} <i>#{__('Continue')}</i> #{__(`, will proceed to the next step, where we'll ask for feedback on your experience.`)}
+
+  a(href='/feedback' role='button' draggable='false') #{__('Continue')}
```
pcraig3: We can't put a continue button on this page. they just have to click off the modal window.

views/offramp/financial.pug
```diff
@@ -24,3 +24,7 @@ block content
     
     a(href='/financial/income' role='button' draggable='false') #{__('Go back')}
 
+    +banner('red')
+        p <strong>#{__('Survey Instructions:')}</strong> #{__('If you are finished, ')} <a href="/offramp/confirmation">#{__('click here')}</a>.
```
pcraig3: I think this should be a mixin so that if we want to update it, we don't do it on every page.

views/offramp/confirmation.pug
nmakuch: I thought the trello card said to link it to the feedback page, but I think the content was actually describing something that happens outside of the prototype

PR #313, opened 2019-10-23
Feature/survey banner

katedee:
## Survey Banner
- a banner that reminds people to imagine that it's their info on the income page, name page, and address page.
- had to add passing the current route so the pug include can confirm what page we're on, and display the banner then (if necessary)
- there are no designs (other than requesting it be fixed to the top of the window), so open to ✨changes✨ 

survey_banner

EDIT: asked to remove the phase banner, here's a screenshot
localhost_3005_personal_name


pcraig3:
We could also put this on the `/login/SIN` page because it says "[Your name here], thanks for your filing code."

PR #319, opened 2019-10-24
Make use of Azure Container Registry instead of DockerHub

charlesmorin:
### This PR consists in
- Changing the `testBuildDeploy.yml` workflow so we make use of Azure Container Registry instead of public DockerHub.
- Adding a couple of Terraform scripts to create an Azure Container Registry and Azure Key Vault.
- Adding instructions to create an Azure Service Principal, which be required when moving from GitHub secrets to Azure Key Vault.
- Making minor wording changes in existing Terraform scripts, docs and variable files.

Next step will be to move from GitHub secrets to Azure Key Vault, so our secrets are stored in a PB environment within CRA Azure Cloud subscription.


scripts/terraform/appservice/terraform.tfvars
```diff
@@ -1,7 +1,7 @@
-location = "canada central"
+location = "Canada Central"
 resource_group_name = "cdscracollab-innovation-rg"
 appservice_plan_id = "alphaPlan"
-appservice_name = "claim-tax-benefits-test"
```
pcraig3: Do we want to remove the '-test'? I thought this was in here so we could set up other app services.

scripts/terraform/azure-container-registry/USAGE.md
```diff
@@ -0,0 +1,34 @@
...
+# What this script does
+
+This script creates a new Azure Container Registry (ACR).
+
+This script can also be used to destroy an Azure Container Registry (ACR).
+
+# Variables
+Variables can be defined in the `terraform.tfvars` file. Defaults are also defined in `create-acr.tf`
+
+# Usage
```
pcraig3: The first step is running `terraform init`, if you haven't already.

scripts/terraform/key-vault/USAGE.md
```diff
@@ -0,0 +1,34 @@
...
+This script creates a new Azure Key Vault so we can store our secrets in it.
+
+This script can also be used to destroy an existing Azure Key Vault.
+
+# Variables
+Variables can be defined in the `terraform.tfvars` file. Defaults are also defined in `create-key-vault.tf`
+
+# Usage
+
+## Create a new Azure Key Vault
```
pcraig3: Similar comment to above: `terraform init`.

scripts/terraform/appservice/terraform.tfvars
charlesmorin: Good catch. We can change it no problem. You are right, the goal was to actually avoid overriding what we have on "prod" now.

pcraig3: Should we do that with the other names too? I mean, I don't think we need to but either works.

PR #332, opened 2019-10-29
Add property tax to check answers page

katedee:
Update spec for checkanswersformat
Update the cypress test and check answers fixture

localhost_3005_checkAnswers (4)

### DIFF
--- a/locales/fr.json
+++ b/locales/fr.json
@@ -435,5 +435,6 @@
   "Sudbury": "Sudbury",
   "Thunder Bay": "Thunder Bay",
   "Windsor": "Windsor",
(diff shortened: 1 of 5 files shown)


locales/fr.json
```diff
@@ -434,5 +434,6 @@
   "St. Catharines-Niagara": "St. Catharines-Niagara",
   "Sudbury": "Sudbury",
   "Thunder Bay": "Thunder Bay",
-  "Windsor": "Windsor"
-}
+  "Windsor": "Windsor",
+  "Did you pay property tax?": "Did you pay property tax?"
```
pcraig3: Let's load the page in french with the other key as well otherwise it will hang around for later prs.

PR #335, opened 2019-10-30
Fix/ca translations

katedee:
### Translate values on check answers data accordingly
- This means translating any strings, but not dollar values
- Also adding locale formatting for date in checkAnswers
- Add locale check to spec for formatAnswers
- add missing keys for checkAnswers that need translation

localhost_3005_checkAnswers


EliseKa:
Can we write "J’accepte" instead of "J'approuve" for the agree button? It will be consistent with the "agree" on the other pages :)

katedee:
@EliseKa of course! done :)

PR #344, opened 2019-11-06
Add blue banner for confirmation code

brdunfield:
Addresses the other half of #340 , implementing the banner shown in https://trello.com/c/i1twNn0Q/371-updates-to-confirmation-page

### DIFF
--- a/public/scss/_base.scss
+++ b/public/scss/_base.scss
@@ -128,6 +128,31 @@ li:not(:last-of-type) {
   }
 }
 
+.confirmation-table {
+  width: 100%;
+  margin-bottom: $space-sm;
+  color: white;
+  background-color: $color-blue-light;
+  border: 3px solid $color-solid-blue;
+  text-align: center;
+  font-size: 24pt;
+  margin-bottom: 1.5em;
+
+  @include sm {
+    width: 80%;
+  }
+
+  thead th {
+    padding-top: 1em;
+    font-weight: normal;
+  }
+
+  tbody td {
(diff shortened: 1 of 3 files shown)


views/confirmation/confirmation.pug
```diff
@@ -8,7 +8,7 @@ block content
   h1 #{title}
 
   div
-    table.pure-table
```
pcraig3: good catch, this was left over from the very early dayy

public/scss/_variables.scss
```diff
@@ -4,6 +4,7 @@ $color-blue-dark: #284162;
 $color-blue-light: #335075;
 $color-banner-blue: #4B98B2;
 $color-banner-blue-light: #DFF8FD;
+$color-solid-blue: #0078FB;
```
pcraig3: I know the `banner` colors are different, but can we say `color-blue-solid` instead? 
It's a weak convention, but let's keep it.

public/scss/_base.scss
```diff
@@ -128,6 +128,31 @@ li:not(:last-of-type) {
   }
 }
 
+.confirmation-table {
+  width: 100%;
+  margin-bottom: $space-sm;
+  color: white;
+  background-color: $color-blue-light;
+  border: 3px solid $color-solid-blue;
```
pcraig3: I think the border looks odd (looks like a focus style). 

But also I don't want to leave code here all day, so let's catch up with @brianhendrick1 when he's back.

PR #345, opened 2019-11-07
Refactor the doYesNo middleware

pcraig3:
Previously we always had "yes" or "no" pages in front of "amount" pages, so I wrote a pretty basic middleware that assumed this would always be the case.

Now that it is no longer the case, we have weird functions that do a bunch of copy+paste stuff and we gotta get rid of that nonsense.

Updated the doYesNo function so that it can handle nulling out more than just the "amount" variable, and it's no longer hardcoded to use the path + "/amount" url that we were assuming before.

The changes from how "doYesNo" worked before are:
- we pass in an array of fields as the second parameter rather than just the "Amount" string
- the redirect isn't hardcoded anymore, instead we're using the "routes" array to pick the next url to go to


pcraig3:
Thanks @nmakuch! I'm gonna merge as-is with my eyes open etc.

PR #350, opened 2019-11-08
CSRF Tokens

brdunfield:
Working on tests now

### DIFF
--- a/app.js
+++ b/app.js
@@ -23,7 +23,7 @@ const express = require('express'),
     currencyFilter,
     isoDateHintText,
   } = require('./utils'),
-  csrf = require('csurf');
+  csrf = require('csurf')
 
 // initialize application.
 var app = express()
--- a/app.js
+++ b/app.js
@@ -22,7 +22,8 @@ const express = require('express'),
     checkLangQuery,
     currencyFilter,
     isoDateHintText,
-  } = require('./utils')
+  } = require('./utils'),
+  csrf = require('csurf');
 
 // initialize application.
 var app = express()
@@ -36,6 +37,20 @@ app.use(express.json())
 app.use(express.urlencoded({ extended: false }))
 app.use(cookieParser(process.env.app_session_secret))
 app.use(require('./config/i18n.config').init)
+
+// CSRF setup
+app.use(
+  csrf({
+    cookie: true,
+    signed: true,
+  }),
+)
+
+// append csrfToken to all responses
+app.use(function(req, res, next) {
+  res.locals.csrfToken = req.csrfToken()
+  next()
+})
 
 // in production we may want to use other than memorysession
 app.use(sessionConfig)
--- a/app.js
+++ b/app.js
@@ -23,7 +23,8 @@ const express = require('express'),
     currencyFilter,
     isoDateHintText,
   } = require('./utils'),
-  csrf = require('csurf')
+  csrf = require('csurf'),
+  cookieConfig = require('./config/cookie.config')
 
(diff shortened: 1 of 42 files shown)


app.js
```diff
@@ -37,6 +38,20 @@ app.use(express.urlencoded({ extended: false }))
 app.use(cookieParser(process.env.app_session_secret))
 app.use(require('./config/i18n.config').init)
 
+// CSRF setup
+app.use(
+  csrf({
+    cookie: true,
+    signed: true,
+  }),
+)
```
pcraig3: Let's stick the generic cookie config that we already have in here as well: https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/config/cookie.config.js

Should just go in cleanly using the spread operator `...cookieConfig`.

app.js
```diff
@@ -37,6 +38,20 @@ app.use(express.urlencoded({ extended: false }))
...
+
+// append csrfToken to all responses
+app.use(function(req, res, next) {
+  res.locals.csrfToken = req.csrfToken()
```
pcraig3: This is done so that the templates have access to the token, right?

views/personal/residence.pug
```diff
@@ -29,4 +29,6 @@ block content
           option(value='Non Resident') #{__('Non Resident')}
     input#redirect(name='redirect', type='hidden', value='/personal/address')
 
+    input(type='hidden', name="_csrf", value=csrfToken)
```
pcraig3: We could put this in the "formButtons" mixin, since every form has a button. 
We could also not because an input is not a button. 

What do you think @katedee?

routes/confirmation/confirmation.spec.js
```diff
@@ -13,37 +14,59 @@ describe('Test confirmation urls', () => {
 })
 
 describe('Test /review', () => {
+  const session = require('supertest-session')
+
+  function extractCsrfToken(res) {
+    var $ = cheerio.load(res.text)
+    return $('[name=_csrf]').val()
+  }
```
pcraig3: Let's put this into a generic file so that we don't copy+paste it everywhere.

Maybe `/utils/index.js`, or a `/routes/utils.js` file — not too bothered where.

routes/confirmation/confirmation.spec.js
```diff
@@ -13,37 +14,59 @@ describe('Test confirmation urls', () => {
...
+  let csrfToken,
+    cookie
+
+  beforeEach(async () => {
+    let testSession = session(app)
+    const getresp = await testSession.get('/financial/income')
+    cookie = getresp.headers['set-cookie']
+    csrfToken = extractCsrfToken(getresp)
+  })
```
pcraig3: Is it possible to wrap this up as well as a generic function? (not sure because it reassigns some variables).

My impression is that we can't, so maybe we just leave it at that.

app.js
brdunfield: Correct

views/personal/residence.pug
brdunfield: I think it might make sense to but it in formButtons, since this input is needed on every form page but only form pages. This would take the guesswork out of remembering whether you need it or not.

brdunfield: I've added it to the mixin for now, I can always revert if we decide it's a bad idea.

katedee: I think putting it there is a good idea, especially since it's on every page. We can always change the name of the mixin if it bothers anyone, but I don't really mind.

PR #351, opened 2019-11-12
Test/xml maker

katedee:
## ✨XML Maker ✨

It's not as sparkly as I make it sound. There's probably some changes I forsee happening either in the short term, or once we're more aware of the format. I've done my best to leave that in the comments as I went, because I knew come PR time it'd all be a blur.

There’s a chance we may need an XML output to file through NETFILE. Without being sure of what that exactly looks like, I took my best guess based off of a .tax file that @pcraig3 got using [Adv Tax](https://www.aclasssoft.com/).

Converted the `.tax` file to a js object, using [xml-js](https://www.npmjs.com/package/xml-js) which can turn json/js objects into xml and vice versa. 

From there it’s just a matter of adding return lines as necessary. It’s pretty rudimentary right now, but it spits out xml, which can then be downloaded via a get on our `/download` page (aka just clicking the button on confirmation)

One big thing I imagine need in the future is adding an options to the dateToLine, which would tell our switch statement what needs to be done with the line, vs going on a line by line case basis (I’ve left some comments in xmlFormat.js talking about some of this)

There are also no dynamic names for the file right now because I didn't want it to spam files when we run cypress tests


routes/confirmation/xmlFormat.js
```diff
@@ -0,0 +1,450 @@
...
+
+        break
+    }
+  })
+  
+
+  const data = convert.json2xml(newXml , {spaces: 2})
+
+  // this because I don't want an output everytime we run the spec files
+  if(createFile) {
```
pcraig3: We can look for `process.env.NODE_ENV !== 'test'` or something like that. 

`npm test` and the `npm run cypress*` commands both run in "test" mode.

xml_output/template-xml.js
```diff
@@ -0,0 +1,81 @@
...
+              "type": "element",
+              "name": "t1:T1Return",
+              "elements": [
+                {
+                  "type": "element",
+                  "name": "comm:SoftwareID",
+                  "elements": [
+                    {
+                      "type": "text",
+                      "text": "024G"
```
pcraig3: Let's change this to `FLCT`

xml_output/template-xml.js
```diff
@@ -0,0 +1,81 @@
...
+                    }
+                  ]
+                },
+                {
+                  "type": "element",
+                  "name": "comm:SoftwareRevisionDate",
+                  "elements": [
+                    {
+                      "type": "text",
+                      "text": "2019-02-22"
```
pcraig3: Let's make this different too — maybe today's date?

routes/confirmation/confirmation.controller.js
```diff
@@ -2,9 +2,26 @@ const { checkSchema } = require('express-validator')
...
+
+    outputXML(req, true)
+  })
+
+  app.get('/download', (req,res) => {
+    /**
+     * I imagine back in outputXML, we'll want to disable the download button until the function is done. We'll rarely see it happen, but just to avoid any issues around downloading and incomplete file
+     */
+    const file = './xml_output/taxfile-2018.xml'
+    res.download(file)
```
pcraig3: If the file isn't there, this throws an error. 

Let's do a `try`/`catch` maybe and just `res.status(500).send('Some generic message')`.

Better than reading out the directory structure.

image

PR #355, opened 2019-11-15
Task/french currency

katedee:
## French Currency 💲

Essentially let's make sure we store values in a consistent format ('210.10'), and then leverage currencyFilter on displaying that back in the appropriate format, based on our current locale/language

| English  | French |
| ------------- | ------------- |
| localhost_3005_financial_income | localhost_3005_financial_income_lang=fr |


katedee:
Currently having issues with the french currency formatting in testing `currencyFilter` in `index.spec.js` fail up here, but have zero issues locally :/

package.json
```diff
@@ -52,6 +52,7 @@
     "express": "~4.17.1",
     "express-session": "^1.16.2",
     "express-validator": "^6.2.0",
+    "full-icu": "^1.3.0",
```
pcraig3: Let's make this a `devDependency`

utils/index.js
```diff
@@ -194,6 +194,17 @@ const doYesNo = (claim, fields) => {
   }
 }
 
+
+const postAmount = (amount, locale) => {
+  if(amount && (locale === 'fr')) {
+    const formattedAmount = amount.replace(',','.').replace(' ', '')
```
pcraig3: It's a bit pedantic, but can we put a space between the args the first replace, just so it's a bit easier to read?

```
amount.replace(',', '.').replace(' ', '')
```

pcraig3: Also, this will only catch the first instance of a thing unless you make it a global replace.

```
"1 1 1 1 1 1 1".replace(' ', '') 
// results in "11 1 1 1 1 1"
```

utils/index.js
```diff
@@ -251,13 +262,22 @@ const hasData = (obj, key, returnVal = false) => {
...
+  const filteredAmount = amount.toLocaleString(localeSetting, {
     minimumFractionDigits: fractionDigits,
     maximumFractionDigits: fractionDigits,
   })
+
+  if(locale === 'fr') {
+    return `${filteredAmount}$`
+  }
+
+  return `$${filteredAmount}`
```
pcraig3: This is the same function as in the cypress tests, right? Can we just use this one or is there a reason to have 2?

(also, let's remove `fractionDigits`)

utils/index.spec.js
```diff
@@ -36,6 +37,42 @@ describe('Test SINFilter', () => {
   })
 })
 
+describe('Test currencyFilter', () => {
```
pcraig3: Let's write a few tests for `postAmount` as well to test the repeating spaces thing.

views/login/questions/trillium.pug
```diff
@@ -4,6 +4,8 @@ block variables
   -var title = __('Ontario Trillium Benefit amount and payment type')
   -var trilliumAmount = hasData(body, 'trilliumAmount') ? body.trilliumAmount : ''
   -var trilliumPaymentMethod = hasData(body, 'trilliumPaymentMethod') ? body.trilliumPaymentMethod : ''
+  -var placeholder= currencyFilter(0, getLocale()).replace('$', '')
+  -var formattedValue= (trilliumAmount !== '') ? currencyFilter(trilliumAmount, getLocale()).replace('$', '') : ''
```
pcraig3: Is there an easy way to make `currencyFilter({amount}, getLocale()).replace('$', '')` into a mixin or function or something? Sucks to have to write it this way so many times.

cypress/utils.js
```diff
@@ -1,4 +1,20 @@
...
+
+  const localeSetting = (locale === 'en') ? 'en-US' : 'fr-CA'
+
+  const filteredAmount = amount.toLocaleString(localeSetting, {
+    minimumFractionDigits: fractionDigits,
+    maximumFractionDigits: fractionDigits,
+  })
+
+  if(locale === 'fr') {
+    return `${filteredAmount}$`
```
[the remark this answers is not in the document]
pcraig3: (just kidding)

views/login/questions/trillium.pug
katedee: I can look into it

utils/index.js
katedee: For whatever reason I could access this from the Cypress tests. Maybe I was just doing something wrong, but it seemed like you can't access outside of the cypress folder :/ 
I'll take another look though, because I also don't like duplicating it.

PR #356, opened 2019-11-18
Register to vote pages

### LINKED ISSUE #339: HOLD: Insert two new pages for voter registration before https://claim-tax-benefits.azurewebsites.net/checkAnswers
Please create two pages:
Page 1 looks like this:
https://projects.invisionapp.com/share/3HULMEV2VWZ#/screens/390966129

Page 2 looks like this with one change: https://projects.invisionapp.com/share/3HULMEV2VWZ#/screens/390966131/comments

"Check the boxes that apply to you:" should be "Check the boxes if:"

Thanks!

brdunfield:
https://trello.com/c/u4d5IN49/382-m-add-election-screens-to-online-tool

Closes #339 

The new pages are ate `/vote/optIn` and `vote/confirmRegistration`
The answers appear on the checkAnswers page


brdunfield:
Still waiting on some content updates for this one, but if either of you have some time @pcraig3 @katedee , I would appreciate code feedback - particularly on the tests. I'm not sure if I've included everything that I need to.

config/routes.config.js
```diff
@@ -43,6 +43,8 @@ const routes = [
   { path: '/trillium/longTermCare/type', editInfo: 'deductions.trilliumLongTermCareTypeClaim' },
   { path: '/trillium/longTermCare/type/amount', editInfo: 'deductions.trilliumLongTermCareAmount' },
   { path: '/deductions/climate-action-incentive' },
+  { path: '/vote/optIn' },
+  { path: '/vote/confirmRegistration', editInfo: 'vote.voterPageEdited' },
```
pcraig3: Is there a way to use another one of the keys instead of this "edited" key?

locales/en.json
```diff
@@ -461,5 +463,25 @@
...
+  "The Canada Elections Act gives rules about elections and voter information. Provincial and territorial governments, members of Parliament, registered political parties, and candidates at election time can have your voter information.": "The Canada Elections Act gives rules about elections and voter information. Provincial and territorial governments, members of Parliament, registered political parties, and candidates at election time can have your voter information.",
+  "If you complete this section, you will be registered to vote until you file your next tax return.": "If you complete this section, you will be registered to vote until you file your next tax return.",
+  "I allow CRA to give my name, address, date of birth and citizenship status to Elections Canada. Elections Canada uses the information to update voter records.": "I allow CRA to give my name, address, date of birth and citizenship status to Elections Canada. Elections Canada uses the information to update voter records.",
+  "Check the boxes if:": "Check the boxes if:",
+  "test": "test",
+  "test2": "test2",
```
pcraig3: Not sure all these keys are getting used. 

I would say:
- delete all the new keys
- run through the pages again in dev mode

Then the "good" keys will be regenerated.

routes/vote/vote.controller.js
```diff
@@ -0,0 +1,53 @@
...
+  app.post(
+    '/vote/confirmRegistration',
+    checkSchema(confirmRegistrationSchema),
+    checkErrors('vote/confirmRegistration'),
+    postConfirmRegistration,
+    doRedirect,
+  )
+}
+
+const postOptIn = (req, res, next) => {
```
pcraig3: Hmm, we have a [`doYesNo`](https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/utils/index.js#L158) middleware that probably works for this use case, (except for the `voterPageEdited` thing, might require changing the key or some code). 

The [commit](https://github.com/cds-snc/cra-claim-tax-benefits/commit/0a48462236c58a46bc1f308be8a9735e062a1bbe) might help you, have a look.

I'd like to keep the custom Post methods to a minimum.

views/vote/confirmRegistration.pug
```diff
@@ -0,0 +1,29 @@
...
+
+  h1 #{title}
+  
+  div
+    p #{__('The Canada Elections Act gives rules about elections and voter information. Provincial and territorial governments, members of Parliament, registered political parties, and candidates at election time can have your voter information.')}
+
+    p #{__('If you complete this section, you will be registered to vote until you file your next tax return.')}
+    
+  form.cra-form(method='post')
+    .multiple-choice.multiple-choice--checkboxes
```
pcraig3: Let's wrap this in a [fieldset](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset) like our other radios mixin: https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/views/_includes/radiosMixin.pug

Then we can remove the `label` below and use a `legend` instead.

views/vote/optIn.pug
```diff
@@ -0,0 +1,20 @@
...
+
+  h1 #{title}
+
+  div
+    p #{__('Would you like to register to vote in the federal election?')}
+
+  form.cra-form(method='post')
+    +radiosYesNo('confirmOptIn', 'Would you like to register to vote in the federal election?', confirmOptIn)
+
+    input#redirect(name='redirect', type='hidden', value='/vote/confirmRegistration')
```
pcraig3: Usually we put the 'No' URL in here instead of the 'Yes' URL. ie, if you say No, you end up redirected to this URL.

config/routes.config.js
brdunfield: I played around with that for awhile - the issue is we want it to show up if either of them have been ticked. Might be possible to do it for whether they ticked yes or no on the previous page?

routes/vote/vote.controller.js
brdunfield: doYesNo seems to specifically target deductions

```req.session.deductions[claim] = false

    // fields is an array
    fields.map(field => {
      // if the field is truthy
      if (req.session.deductions[field]) {
        // "Amount" fields are zeroed -- others are nulled
        if (field.endsWith('Amount')) {
          if (Object.keys(req.session.deductions[field]).includes('amount')) {
            req.session.deductions[field].amount = 0.0
          } else {
            req.session.deductions[field] = 0
          }
        } else {
          req.session.deductions[field] = null
        }
      }
    })```

Which is why I wrote a new one. We could look into refactoring?

File: config/routes.config.js (hunk 1 above)
Comment (brdunfield) [reply config/routes.config.js]: The existing code doesn't support using the Yes / No from the optIn page to trigger this - it's just looking for a non-null value (which it will be even if set to 'No'). We can either refactor to support this use case, or keep it the way it is in this PR.

Comment (pcraig3) [reply config/routes.config.js]: What about if we only asked the first one? 
(If they're not a Canadian citizen, we don't let them vote, right?)

Comment (brdunfield) [reply config/routes.config.js]: Both questions need to be answered yes for them to be registered, if they only tick the second one we still want this page to show up.

File: routes/vote/vote.controller.js (hunk 3 above)
Comment (pcraig3) [reply routes/vote/vote.controller.js]: Yeah, we should refactor.

We could do `field.endsWith('Amount') || field.endsWith('Edited')`, since it sets it to `0`.

File: config/routes.config.js (hunk 1 above)
Comment (pcraig3) [reply config/routes.config.js]: I think we're pretty deep in edge-case land, but I defer to your judgement since it seems pretty low-effort overall.

File: locales/en.json [hunk 6]
```diff
@@ -464,24 +464,18 @@
   "Public or non-profit long-term care home": "Public or non-profit long-term care home",
   "Did you pay property tax?": "Did you pay property tax?",
   "Property tax paid": "Property tax paid",
+  "Invalid value": "Invalid value",
+  "$999.00": "$999.00",
+  "$0.00": "$0.00",
```
pcraig3: Hmm, do you know where these three values are coming from? They all look like they shouldn't be here to me.

routes/vote/vote.controller.js
```diff
@@ -29,19 +29,19 @@ const postOptIn = (req, res, next) => {
   req.session.vote.confirmOptIn = confirmOptIn
 
   // if yes, go to second page of vote
-  // if no, go to confirmation
-  if (confirmOptIn == 'No') {
-    req.session.vote.voterCitizen = null
-    req.session.vote.voterConsent = null
-    req.session.vote.voterPageEdited = 0
-    return res.redirect('/checkAnswers')
+  if (confirmOptIn == 'Yes') {
```
pcraig3: Let's use strict equals (`===`) all the time so that we don't end up with weird casting problems down the road.

routes/vote/vote.controller.js
pcraig3: Oh, except the `deductions` problem. 

Okay, let's leave this as-is for now.

config/routes.config.js
brdunfield: I agree, this is fairly edge case. If it becomes a problem we can revisit. Changed to voterCitizen

PR #359, opened 2019-11-19
Add Cypress full run with amounts

katedee:
## Cypress tests, with amounts!
Another Cypress test to run through the app, and say yes to questions and add amounts. This includes entering an amount, navigating one page forward, and then one page back to check the amount is there.

## Why do things cache?
Only issue is it seems to cache the input entered between runs. Nobody knows why.

~In order for a repeat run to work, I have to kill cypress, and re-run for a clean run--looking into this, but does anyone else want to take a crack at it?~

New code doesn't expect amounts to be empty. Just clears whatever's in the field and enters a new amount. Easy.

### DIFF
--- a/cypress/integration/full_run_with_amounts.spec.js
+++ b/cypress/integration/full_run_with_amounts.spec.js
@@ -5,7 +5,7 @@ const {
   getAddress,
 } = require('../utils.js')
 
-describe('Full run through', function() {
+describe('Full run through saying "yes" to everything', function() {
   before(function() {
     cy.visit('/clear')
     cy.visit('/')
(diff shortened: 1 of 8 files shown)


cypress/integration/full_run_with_amounts.spec.js
```diff
@@ -0,0 +1,286 @@
+const {
+  checkTableRows,
+  allIncomeRows,
+  getBenefitsBreakdownRows,
+  getAddress,
+} = require('../utils.js')
+
+describe('Full run through', function() {
```
brdunfield: I might change the name of the second suite to something more like "Full run through with amounts", so it's immediately obvious which suite is which.

PR #360, opened 2019-11-20
Landing page update

nmakuch:
- The card was based off of this [trello card](https://trello.com/c/oT9KrXCh)

### Updating the landing page layout + content
- Only real thing to note is that the start button has been added to the middle of the page layout and margins were adjusted to make this look nice
- Some new keys added to the en.locale
- Fr side is broken now though 🤷🏽‍♂️will need to switch that in when we have it

### Before/After

| Before | After |
|--------|-------|
| Screen Shot 2019-11-20 at 11 16 12  |  Screen Shot 2019-11-20 at 11 15 59  |


locales/en.json
```diff
@@ -461,5 +460,15 @@
...
+  "start.p2.2": "under $12,000",
+  "start.p2.3": ". Income includes:",
+  "start.p3.1": "You ",
+  "start.p3.2": "cannot",
+  "start.p3.3": " use this service if you had income from your own business or self-employment.",
+  "start.p4": "You will need:",
+  "start.p5": "Your must answer extra security questions if you want your notice of assessment immediately after filing. You need information such as previous addresses and banking details to answer these questions.",
+  "start.h2-1": "Before you start",
+  "start.h2-2": "Help using the service",
+  "start.p6": "Contact CRA by:"
```
pcraig3: Hey nick, can you group together all the ones that are `start.`? (ie, put them at the top)

locales/fr.json
```diff
@@ -438,5 +438,16 @@
...
+  "start.p2.2": "start.p2.2",
+  "start.p2.3": "start.p2.3",
+  "start.p3.1": "start.p3.1",
+  "start.p3.2": "start.p3.2",
+  "start.p3.3": "start.p3.3",
+  "start.h2-1": "start.h2-1",
+  "start.p4": "start.p4",
+  "start.p5": "start.p5",
+  "start.h2-2": "start.h2-2",
+  "start.p6": "start.p6"
```
pcraig3: Same here, keep them together.

views/start/index.pug
```diff
@@ -5,20 +5,32 @@ block content
...
-      li #{__('start.li4.1')} <i>#{__('service.name')}</i>#{__('start.li4.2')}
+      li #{__('start.li4.1')} <i>"#{__('service.name')}"</i>
       li #{__('start.li5')}
       li #{__('start.li6')}
 
-  div
-    a(href='/login/code' role='button' draggable='false') #{__('button.start')} <span aria-hidden="true">→</span>
+    p #{__('start.p5')}
+
+    h2 #{__('start.h2-2')}
```
pcraig3: lol, yeah `h3` would be misleading. Carry on.

nmakuch: ya I felt like the font size should be a bit smaller but h2 is better for semantics so oh well

locales/en.json
```diff
@@ -39,19 +39,18 @@
   "https://www.canada.ca/en/revenue-agency/services/e-services/e-services-individuals/account-individuals.html?utm_campaign=not-applicable&utm_medium=vanity-url&utm_source=canada-ca_my-cra-account": "https://www.canada.ca/en/revenue-agency/services/e-services/e-services-individuals/account-individuals.html?utm_campaign=not-applicable&utm_medium=vanity-url&utm_source=canada-ca_my-cra-account",
   "canada.ca/taxes-help": "canada.ca/taxes-help",
   "canada.ca/my-cra-account": "canada.ca/my-cra-account",
-  "service.name": "Claim Tax Benefits",
+  "service.name": "Claim benefits by filing taxes",
   "start.banner.1": "The Canada Revenue Agency (CRA) sent you a letter about the new",
   "start.banner.2": " service. ",
   "start.banner.3": "You need information on that letter to file your 2018 taxes with this service.",
```
pcraig3: Also we can delete these now that there's no banner. (in the french version as well)

locales/en.json
```diff
@@ -39,19 +39,18 @@
   "https://www.canada.ca/en/revenue-agency/services/e-services/e-services-individuals/account-individuals.html?utm_campaign=not-applicable&utm_medium=vanity-url&utm_source=canada-ca_my-cra-account": "https://www.canada.ca/en/revenue-agency/services/e-services/e-services-individuals/account-individuals.html?utm_campaign=not-applicable&utm_medium=vanity-url&utm_source=canada-ca_my-cra-account",
   "canada.ca/taxes-help": "canada.ca/taxes-help",
   "canada.ca/my-cra-account": "canada.ca/my-cra-account",
-  "service.name": "Claim Tax Benefits",
+  "service.name": "Claim benefits by filing taxes",
```
pcraig3: Wait, this changes the text in a bunch of places in the app (check the next page for example). Is this the intention?

views/start/index.pug
pcraig3: No, I mean that we use key names like `p2` for paragraphs, but `h3` is an actual HTML element, so it would be weird to say `h2 #{__('start.h3')}`, even though it is the third header element on the page.

All I mean is that the key name and the HTML element are similar.

PR #362, opened 2019-11-22
Add a database for storing access codes

charlesmorin:
This PR consists in adding a terraform script to provision a managed Azure PostgreSQL database server and database to store access codes.
Trello card: https://trello.com/c/7z4b3s6E

Database server and database have already been created in Azure. Passwords for the server admin and user database are stored in the Azure Key Vault under our resource group.


scripts/terraform/azure/postgresql/init-database.sql
```diff
@@ -0,0 +1,31 @@
+CREATE ROLE ctb_user WITH
+	LOGIN
+	NOSUPERUSER
+	NOCREATEDB
+	NOCREATEROLE
+	INHERIT
+	NOREPLICATION
+	CONNECTION LIMIT -1
+	PASSWORD '--> ENTER PASSWORD HERE <--';
```
pcraig3: This is for the user (ie, me or whoever is on the command line) to fill in once the server has been created, right?

scripts/terraform/azure/postgresql/init-database.sql
```diff
@@ -0,0 +1,31 @@
+CREATE ROLE ctb_user WITH
```
pcraig3: This file is to be run after the PG server has been created, right?

scripts/terraform/azure/postgresql/init-database.sql
charlesmorin: Yes, this is to actually create a database user so it can connect to the database.

That could have been done by Terraform as well. However, I didn't see any value of doing it because PostgreSQL for PostgreSQL, better stay with the SQL language than switching between Azure resource template and AWS RDS equivalent. I focused on creating the resources from Terraform, and the post-creation is done through SQL scripts.

scripts/terraform/azure/postgresql/init-database.sql
charlesmorin: Correct. There are multiple way of doing it, but as I said in my previous comment, I wanted to stick with native SQL language instead of coupling myself with vendor-specific post-database-creation workflows. This entire script could be more sophisticated, but since we're moving to AWS, I didn't spend too much time on the Azure one.

PR #368, opened 2019-11-25
Pre security screening page

nmakuch:
### Creating the `/login/notice` page for pre-security question page screening
- Validation testing showed us that we need a page before the security questions if placed at the end of the flow as a break point between full and lite mode, otherwise the task is confusing
- Route was created
- login controller and schema we modified for this new page
- tests were updated
- french is missing

### Note 
I didn't add it to the flow, I figured it would be easier to do once the questions have been moved to the end of the flow, otherwise it would skip to a weird page and not really make sense

### GIF
notice

### DIFF
--- a/routes/login/login.controller.js
+++ b/routes/login/login.controller.js
@@ -44,7 +44,12 @@ module.exports = function(app) {
     '/login/notice',
     checkSchema(noticeSchema),
     checkErrors('login/notice'),
-    postNotice,
+    (req, res, next) => {
+      if (req.body.noticeOfAssessment === 'No') {
+        return res.redirect('/checkAnswers')
+      }
+      next()
+    },
     doRedirect,
   )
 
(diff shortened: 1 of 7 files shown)


routes/login/login.controller.js
```diff
@@ -204,3 +214,14 @@ const postSecurityQuestion = async (req, res) => {
...
 }
+
+const postNotice = (req, res, next) => {
+  const noticeOfAssessment = req.body.noticeOfAssessment
+
+  if (noticeOfAssessment === 'No') {
+    return res.redirect('/')
+  }
+
+  req.session.login.noticeOfAssessment = true
```
pcraig3: Let's not save this. It's not in our user.json

views/login/notice.pug
```diff
@@ -0,0 +1,20 @@
...
+  -var title = __('Notice of assessment')
+  -var noticeOfAssessment = !hasData(data, 'login.noticeOfAssessment') ? '' : data.login.noticeOfAssessment ? 'Yes' : 'No'
+
+block content
+
+  h1 #{title}
+
+  p #{__('CRA sends your notice of assessment about 2 weeks after you file taxes.')}
+  p #{__('You can download the notice immediately after filing by answering security questions. You will be able to print or save the downloaded notice.')}
+  p #{__('The questions ask for information such as your previous addresses and banking details. You may need to check your records to find the answers.')}
```
pcraig3: We should tell them that they don't need to answer the security questions to file their taxes, as it's not explicit.

PR #371, opened 2019-11-26
Access code flow changes

brdunfield:
Closes https://trello.com/c/vOZOZpTV/373-m-rejig-the-access-code-sin-dob-logic-so-that-its-closer-to-what-we-want-by-the-end

Essentially, this PR simulates the authentication flow described in the trello card. db.json simulates the new database containing login info (access code, sin, dateOfBirth). This moves main session population to the dateOfBirth post action. Essentially session creation is still done in the access code post, but most fields are left empty until after the correct SIN and Date of Birth are submitted. 

--

Now there are a bunch of different flows we have to account for depending on if someone has skipped a step or gotten one wrong. We also don't want to tell people that their SIN or Date of Birth is wrong specifically, so any mistake on either of those pages will redirect back to the `/login/sin` page with a generic error message. 

It's bad for users, but it's harder to brute force.

Here's a handy table which tries to explains it.

```
| code | sin | dob | redirect       |
|------|-----|-----|----------------|
| -    | ✅  | ✅  | /login/code    |
| -    | -   | ✅  | /login/code    |
| ✅   | ❌  | ❌  | /login/sin     |
| ✅   | ❌  | ✅  | /login/sin     |
| ✅   | ✅  | ❌  | /login/sin     |
| ✅   | -   | ✅  | /login/sin     |
| ✅   | ✅  | ✅  | /personal/name |
```

I've created a new test file `login.session.spec.js` to keep almost all of the session-specific unit tests in for this new flow, so check in there for the scenarios that are covered.


routes/login/login.controller.js
```diff
@@ -212,21 +215,24 @@ const postDateOfBirth = async (req, res, next) => {
...
+  req.session.personal.dateOfBirth = _toISOFormat(req.body)
+
+  // check access code + SIN + DoB
+  // if session doesn't have a sin, throw error
+  if(!req.session.personal.sin) {
+    // error no sin
+    let errObj = {
+      sin: {
+        msg: 'errors.login.matchingSIN',
+      },
```
pcraig3: Let's simplify this. 

Since we're not using the "matching DoB login" anymore, we can add [the `checkErrors` call](https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/routes/login/login.controller.js#L49) as one middleware, and then in our second middleware we can make sure they all add up.

If we we find an error (sin is missing, SIN is wrong, Dob is wrong, etc), we should redirect back to the `/sin` page with an error message.  

As I am writing this, I am realizing that that would mean the url would be wrong, so maybe we have to check the session in the SIN post function.

api/index.js
```diff
@@ -19,4 +20,41 @@ var API = (function(userFound) {
   }
 })(user)
 
-module.exports = API
+var DB = (() => {
```
pcraig3: Could we create a `/db` folder instead so the imports don't change everywhere or now that you've done the work is that a dumb suggestion?

brdunfield: If you think it makes it more maintainable I think it would be worth doing.

routes/login/login.controller.js
```diff
@@ -180,24 +188,19 @@ const postLoginCode = async (req, res, next) => {
...
 
-  if (!user) {
+  if (!validCode) {
     throw new Error(`[POST ${req.path}] user not found for access code "${req.body.code}"`)
   }
 
+  // populate the session with empty variables in the format of user.json
   // setting req.session = {obj} causes an error, so assign the keys one at a time
-  Object.keys(user).map(key => (req.session[key] = user[key]))
+  Object.keys(user).map(key => (req.session[key] = {}))
```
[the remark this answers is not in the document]
brdunfield: I think a req.session.db *could* work. The reason I took this approach is that there are issues where, for example, the sin page on its post call was already expecting the sin to be in the session, which I don't think it would be in the new flow. Initially we're just validating the Access code, and not pulling the rest of the info out of the DB until the DoB call, right?

I guess, in my head, the flow was - check if access code exists, add it to session. Add sin and DoB to session as they are input, validate them in DoB post, and retrieve the rest of the data. (We could re-write the SIN / DoB from that call as well if needed - addressing one of your other comments)

brdunfield:
Had to go catch a bus but wanted to leave next steps here.

I still feel like we should wait until the date of birth post to actually populate from user.json. The next step from my last commit would be to keep the session (aside from session.db) completely empty until then, not the object with empty values that I had previously.

With the addition of the session.db variable, neither the sin or DoB pages will need to depend on the rest of the session existing, so any leftover code there can port over to checking against session.db.

I didn't get a chance to gut any of the schema on these pages yet.

api/index.js
```diff
@@ -19,4 +20,41 @@ var API = (function(userFound) {
...
+    })
+    // return the whole row for now for populating session with initial data
+    return dbUser
+  }
+
+  const validateUser = (login) => {
+    // validates user code / sin / DoB
+    let dbUser = false
+    
+    dbUser = db.find(function(u,i) {
```
katedee: This is a minor thing, but could we explicitly make it `user` instead of `u`. It also looks like we're not using the index  (i) anywhere (in this find and also not in the find on line 30), so no need to pass it. Also for consistency, how do you feel about changing this to use the arrow function (unless there's reason not to)? So:
```
db.find( user => {
  //do stuff here
})
```

routes/confirmation/xmlFormat.spec.js
```diff
@@ -4,27 +4,27 @@ const { outputXML } = require('./xmlFormat')
 const initialSession = API.getUser('A5G98S4K1')
 
 const dataToLine = [
-  { 
+  {
```
pcraig3: There are no code changes to this file, it's just formatting.

utils/index.js
```diff
@@ -297,6 +319,26 @@ const currencyFilter = (number, locale = 'en') => {
...
+
+  const dateParts = date.split('-')
+
+  if (dateParts.length !== 3 || dateParts[2].length > 2) {
+    throw new Error(`[GET /login/dateOfBirth] Bad date "${date}": must be formatted yyyy-mm-dd`)
+  }
+
+  return `${dateParts[2]} ${dateParts[1]} ${dateParts[0]}`
+}
+
```
pcraig3: I moved this method up here from lower down since it's a pug filter. Didn't change the code.

pcraig3:
Note: LGTM is mad about there not being rate-limiting. Looks like something we can add without too much bother, but let's save that for another PR.

https://lgtm.com/rules/1506065727959/

api/index.js
pcraig3: yup 👍

utils/index.js
```diff
@@ -132,11 +132,22 @@ const doRedirect = (req, res) => {
 }
 
 // Render a passed-in template and pass in session data under the "data" key
-const renderWithData = template => {
+const renderWithData = (template, { errorsKey } = {}) => {
   return (req, res) => {
-    res.render(template, {
+    let errors = undefined
+
+    // if there are errors in the session under the specified key, add them to the template
```
brdunfield: This is to allow us to render errors from another page? (in this case - render DoB errors on the SIN page after the redirect?)

routes/login/login.session.spec.js
```diff
@@ -0,0 +1,197 @@
+const session = require('supertest-session')
```
brdunfield: Making this its own file was a good idea 👍

routes/login/login.session.spec.js
```diff
@@ -0,0 +1,197 @@
...
+describe('Test /login SESSION responses', () => {
+  beforeEach(async () => {
+    testSession = session(app)
+    const getresp = await testSession.get('/login/code')
+    cookie = getresp.headers['set-cookie']
+    csrfToken = extractCsrfToken(getresp)
+  })
+
+  // TEST DONT RETURN TO ACCESS CODE PAGE IF ONLY SIN (NO ACCESS CODE OR DOB)
+  test('it returns a 200 on /login/dob page when only the SIN exists', async () => {
```
brdunfield: We might be able to reduce some of the code here by using couple maps for requests that do and don't lead to various places,  like we do here in login.spec:

```badDoBRequests.map(badRequest => {
        test(`it returns a 422 with: "${badRequest.label}"`, async () => {
          const response = await authSession
            .post('/login/dateOfBirth')
            .use(withCSRF(cookie, csrfToken))
            .send({ ...badRequest.send })
          expect(response.statusCode).toBe(422)
        })
      })```

utils/index.js
pcraig3: Yep, that's what it's for. Only being used by /login/code, and /login/sin

routes/login/login.session.spec.js
pcraig3: Yeah, it was getting really complex :/

routes/login/login.session.spec.js
pcraig3: We could do, but generally the more abstract the tests are, the harder they are to read. I was talking with @nmakuch a few days ago about some of the tests that were being mapped to explain how they worked.

For these ones, I am kind of okay with the copy+paste but if we wanted to make it less lines, I think checking the responses (for errors, etc) would save a bunch of copy + paste.

brdunfield: Fair enough, they made sense to me but I can see how they'd be abstract enough to be not worth the code savings. 👍

brdunfield:
Turns out I can't add a review because I made the PR, but LGTM! 🦑

PR #375, opened 2019-11-26
Added quotes around terraform vars

pcraig3:
Our linter was mad, so adding the vars seemed like the thing to make it happy.


pcraig3:
🤑

PR #376, opened 2019-11-27
Merge AWS Terraform scripts to master

charlesmorin:
This PR consists in merging all work that was done towards AWS (no impact on actual codebase) so we avoid having to rebase all the time with the `master` branch.

AWS work is on hold right now as our main focus is the Concierge testing.


pcraig3:
Okay, so I am okay with merging this but I want to get to the root of the difference between our setups. 

When I have my VS Code Terraform plugin disabled, I don't see any issues (or syntax highlighting).

Screen Shot 2019-11-27 at 2 02 55 PM

When it's enabled, I do see issues (I think about the quotes issue).

Screen Shot 2019-11-27 at 1 41 08 PM

There's also something telling me I should download "experimental language server". Should I try turning that on? Did you do it?

Screen Shot 2019-11-27 at 1 56 01 PM

charlesmorin:
> Okay, so I am okay with merging this but I want to get to the root of the difference between our setups.
> 
> When I have my VS Code Terraform plugin disabled, I don't see any issues (or syntax highlighting).
> 
> Screen Shot 2019-11-27 at 2 02 55 PM
> 
> When it's enabled, I do see issues (I think about the quotes issue).
> 
> Screen Shot 2019-11-27 at 1 41 08 PM
> 
> There's also something telling me I should download "experimental language server". Should I try turning that on? Did you do it?
> 
> Screen Shot 2019-11-27 at 1 56 01 PM

I believe we should focus on the official Terraform validation command: `terraform validate`. This command will essentially lint and confirm that the script is syntactically correct. Reason why I am saying this is because the VScode extension is not an official one, but an extension that was created by a single person. I would take the red squiggly with a grain of salt. That's a conversation I had with Max Neuvians as well in the early development of the AWS pivot, and he seemed to be confused by the IDE behavior as well.

maxneuvians:
Assume that if `terraform validate` passes you are good to go

pcraig3:
image

PR #378, opened 2019-11-28
Upgrade dependencies

pcraig3:
Only breaking change is the nodemon library, but I've tested it and it works fine.

Process for upgrades is:

- `npm outdated`
- (Update deps to latest in `package.json`)
- `rm -rf node_modules package-lock.json`
- `npm i`
- `npm test`
- `npm run lint`
- `npm run cypress:cli`
- and then boot up the app and click through a few pages

If all that works (it did), then we're good.

This PR closes #377 since it includes the same upgrade (and others)


pcraig3:
Thanks @brdunfield !

PR #379, opened 2019-11-28
Add express-rate-limiter

brdunfield:
Seeing a weird cypress bug locally, keeping this a draft PR until I get it sorted

The [access code flow](https://github.com/cds-snc/cra-claim-tax-benefits/pull/371) PR introduced a post handler that LGTM flagged as being expensive, and suggested we add rate limiting. Added express-rate-limiter to the offending page to reduce our vulnerability to a denial of service attack.

https://trello.com/c/Nl5EgArp/452-s-add-rate-limiting-to-make-lgtm-happy


app.js
```diff
@@ -55,6 +56,14 @@ app.use(function(req, res, next) {
   next()
 })
 
+// set up rate limiter: maximum of five requests per minute
+var limiter = new rateLimit({
+  windowMs: 1*60*1000, // 1 minute
+  max: 120,
+})
+// apply rate limiter to expensive request page(s) - just the one for now
+app.use('/login/dateOfBirth', limiter)
```
pcraig3: I was thinking to pass this in as a middleware in the /login route itself. That might solve your issue?

https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/routes/login/login.controller.js#L43

```js
  app.post(
    '/login/dateOfBirth',
    limiter, // pass in here
    checkSchema(dobSchema),
    checkErrors('login/dateOfBirth'),
    postLogin,
    doRedirect,
  )
```

brdunfield: The issue has mysteriously gone away. It was also on the checkAnswers page and about the Marital Status row, so it seemed extremely unrelated to begin with.

Are there advantages to changing this to a passed middleware?

brdunfield: It looks like cypress works on the first runthrough after a code change, then fails every time after.

brdunfield:
Spoke to Kate and this is probably just the same issue she noticed in her recent PR where cypress doesn't want to run locally multiple times. I think I'm good to merge.

app.js
pcraig3: The advantage would be keeping the middleware with the code itself but this is fine.

PR #380, opened 2019-11-28
Reuse unit tests for amount and year pages

pcraig3:
Allows us to remove a bunch of lines from the bottom of this file.

### DIFF
--- a/routes/login/login.spec.js
+++ b/routes/login/login.spec.js
@@ -650,7 +650,7 @@ questionsAmounts.map(amountResponse => {
       })
     })
 
-    test('it returns a 422 response for a good amount but NO payment method', async () => {
+    test('it returns a 422 response for a good amount but NO year', async () => {
       const response = await request(app)
         .post(amountResponse.url)
         .use(withCSRF(cookie, csrfToken))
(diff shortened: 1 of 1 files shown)


pcraig3:
Robert Galanakis: 

> “The fastest code is the code which does not run. The code easiest to maintain is the code that was never written.”

routes/login/login.spec.js
```diff
@@ -653,24 +665,8 @@ questionsAmounts.map(amountResponse => {
...
-            taxReturnYear: '2018',
-            taxReturnAmount: goodAmount,
-            redirect: '/start',
-          })
-        expect(response.statusCode).toBe(302)
-        expect(response.headers.location).toEqual('/start')
-      })
-    })
-
-    const goodYears = ['2018', '2017']
```
brdunfield: Was there a reason we originally tested against 2 years?

pcraig3: It was rewritten from an earlier check where only 2 values were accepted.

We accept tons of year values so these two are arbitrary.

pcraig3: Okay, you win.

pcraig3:
image

PR #383, opened 2019-11-29
Check if SIN/DoB match the provided access code

brdunfield:
https://trello.com/c/i3DdANSc/454-s-sin-and-dob-might-be-for-a-different-a-different-code

This PR uses the following flow:
- Check if access code is valid on post of /login/code
- on /login/dateOfBirth post, checks if the SIN and DoB match a user.
- THEN, check if that user is the same user as the previously validated code. If not, redirect back to /login/code, with an error. (Will need to have content look at this but I just threw in a placeholder for now)

Notable code things:
- Removed some schema and tests around checking if the access code was within a hardcoded array consisting of only the old code in upper / lowercase. DB validation should replace this, so it's no longer needed.
- I'm not sure if I'm happy with how the DB methods now return an object with an error prop when something in this new flow fails. It's a pretty different pattern than what we had before.. then again we didn't have a DB before so it makes sense that we'd have a new way of doing things. Feedback on this would be appreciated!

### DIFF
--- a/db/index.js
+++ b/db/index.js
@@ -6,7 +6,7 @@ var DB = (() => {
     code = code.toUpperCase()
 
     // Real DB code will go here
-    return db.find(user => user.code === code) || null
+    return db.find(user => user.code === code) || {"error":  'errors.login.code'}
   }
 
   const validateUser = ({ code, sin, dateOfBirth }) => {
@@ -14,9 +14,14 @@ var DB = (() => {
     sin = cleanSIN(sin)
 
     let row = db.find(user => {
-      if (user.code === code && user.sin === sin && user.dateOfBirth === dateOfBirth) return user
+      if (user.sin === sin && user.dateOfBirth === dateOfBirth) {
+        return user
+      }
     })
-
+    if (row && row.code !== code) {
+      // code from the returned user doesn't match
+      return {"error": "errors.login.codeMatch"}
+    }
     return row || null
   }
 
--- a/db/index.js
+++ b/db/index.js
@@ -6,7 +6,7 @@ var DB = (() => {
     code = code.toUpperCase()
 
     // Real DB code will go here
-    return db.find(user => user.code === code) || {"error":  'errors.login.code'}
+    return db.find(user => user.code === code) || null
   }
 
   const validateUser = ({ code, sin, dateOfBirth }) => {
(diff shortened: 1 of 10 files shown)


db/index.js
```diff
@@ -6,17 +6,22 @@ var DB = (() => {
...
     let row = db.find(user => {
-      if (user.code === code && user.sin === sin && user.dateOfBirth === dateOfBirth) return user
+      if (user.sin === sin && user.dateOfBirth === dateOfBirth) {
+        return user
+      }
     })
-
+    if (row && row.code !== code) {
+      // code from the returned user doesn't match
+      return {"error": "errors.login.codeMatch"}
```
pcraig3: Yeah, this is what I ran into as well, having to return potentially 3 different values:

- `null`: didn't find anything
- `row`: found a match
- `??`: sin and DoB match a given user but the code is wrong

In line with my earlier comment, I would prefer not to return an error string here. 

In which case, 
 
- we have 3 return values
- we can create another function
- if "validateUser" returns null, we can see if "validateCode" returns a row, and (if it does), we can do a logic check in the route for the session DoB plus the session SIN in `postLogin`.

db/index.js
```diff
@@ -6,17 +6,22 @@ var DB = (() => {
     code = code.toUpperCase()
 
     // Real DB code will go here
-    return db.find(user => user.code === code) || null
+    return db.find(user => user.code === code) || {"error":  'errors.login.code'}
```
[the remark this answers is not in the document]
brdunfield: Not a pain, this was what I was referring to above when I said I wasn't totally happy with the implementation 👍 

We do need to distinguish between null and an error though, at least in validateUser - null means it wasn't found, whereas the error means the user exists but the code is wrong.

db/index.js
brdunfield: Oh I typed my earlier comment before I saw this

brdunfield: 👍 This looks perfect, will do

pcraig3: Just to be clear, those last three bullet points were "either/or"s. Doing any one of those would work.

brdunfield:
I opted to just revert validateCode to either a row or null return, and add a third return value for validateUser (which is just a boolean error flag on an otherwise empty object).

I felt like the other options would require more expensive operations to occur, such as potentially more db calls.

PR #389, opened 2019-12-03
Feature/two security questions

katedee:
Basic flow for answering two security questions. 
Essentially we post if the answer info is correct or not as a Bool, alongside the url (I kept the url because we might want to later check _what_ is answered)
We go thru a securityQuestionsRedirect, which: 
- directs back to the questions page if only one security question has been attempted
- direct to checkAnswers if less than two questions have been answered correctly
- carries on with the usual redirect in doRedirect if two questions are correct

There are a couple improvements I'd like, but as we're not sure if we'll be keeping this feature, it stands as is. Most notably what's missing right now:
- any real validation on security questions (and so we post 'true' for each answer at the moment)
- since we post true always, there's no real way to write some meaningful tests right now, so it's without tests, but mostly what we have in place kind of covers it enough at the moment (what's displayed/not display)
- cypress amounts tests is broken atm. I'm not sure what's causing it as when I try to manually reproduce this, there's no issues. So I suspect it's something to do with cypress caching.
- later we might want to check + overwrite for questions already answered in the securityQuestion array, but I'm not overly concerned about it right now


utils/index.js
```diff
@@ -131,6 +131,20 @@ const doRedirect = (req, res) => {
...
+
+  if (securityQuestionValues.length < 2) {
+    //TODO: redirect to the page they came from
+    return res.redirect('/login/securityQuestion')
+  } else if (securityQuestionValues.filter(v => v[0]).length < 2) {
+    //redirect to lite app
+    return res.redirect('/checkAnswers') 
+  }
+
+  next ()
```
pcraig3: Is this valid? I would have thought that `next()` works, but `next ()` wouldn't.

katedee: I should fix it anyways, because it is a bit strange

pcraig3:
This is what I'm getting for the cypress tests that fail for the "amounts" suite:

image

pcraig3:
but it's working on the heroku deploy, so not sure what's the issue.

katedee:
@pcraig3 yes, I'm aware of the error issue in cypress, and explicitly mention it above and posting in channel. I'm not sure what the issue is in the test

pcraig3:
Yup, just thought you didn't have a good error message. 
However trying it again, it appears to be working for me. 

Not sure what the github actions workflow issue is here though. That's new.

brdunfield:
I'm not sure what the status of doing more work on this is, but there's a pretty significant bug which is if the user navigates back past the securityQuestions page and then navigates forward, they are presented with the security questions page again, with the ones they've already answered missing, and no ability to progress again unless they click "skip security questions". (They get an error about needing to select a question).

pcraig3:
I am closing this :gavel:

PR #395, opened 2019-12-05
Add Cypress tests back

pcraig3:
Yesterday, I removed the Cypress tests from the application in Github actions, but that was for expedience. Really, we should put them back because they are making sure our page flow works and they are doing a11y checks for us.

Source: #391 

Found the official cypress github action and am running that one now

Looks way more legit + maintained: https://github.com/cypress-io/github-action

Read the README for setup instructions.

### DIFF
--- a/.github/workflows/testBuildDeploy.yml
+++ b/.github/workflows/testBuildDeploy.yml
@@ -2,21 +2,25 @@ name: Test, build, deploy
 on: push
 jobs:
   test:
-    name: Run tests
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x]
+    name: Run tests
     steps:
       - uses: actions/checkout@master
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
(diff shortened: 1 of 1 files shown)


brdunfield:
Just looking at this PR, all the github actions in the checks are doubled. Is that an artifact of the force pushing? I haven't seen that before. Also, out of curiosity, why are we now testing on multiple versions of node?

pcraig3:
🤷‍♀ because it works, basically.

To give you a better answer though, [there are two currently active Long-Term Support versions of node](https://nodejs.org/en/about/releases/): v10 and v12.


I am using node `10.x` on my local machine as my default, and on my CRA laptop, it's running node `10.13.0` or something (and I can't upgrade it or anything). So it needs to work on node 10.

However, we should also be keeping it up to date with node `12.x` because that's the latest version. 

If any new packages were to break running on either version, we should know about it before we merge.

pcraig3:
> Just looking at this PR, all the github actions in the checks are doubled. Is that an artifact of the force pushing? I haven't seen that before. 

Yeah, I think it's a weird thing that happened because I pushed / force-pushed a bunch of commits to it. Shouldn't happen on other PRs.

PR #396, opened 2019-12-05
Add Ontario property tax page and change flow for OTB

brdunfield:
https://trello.com/c/9Ghd8mMo/465-m-update-otb-property-tax-flow

Added a new page for Ontario property tax, and re-ordered the flow to match the content document.

Jest tests solution:

I've added a new test suite to handle pages that redirect to `/ontario` pages, and separated them out. The original test suite will continue to be used for testing the `/ontario` pages. For now, pages that haven't been updated yet will continue to be in the original test suite, but as they are updated, they should be moved to the new test block, and their URL in the original should be replaced with the `/ontario` page in their flow.


brdunfield:
I am not sure why SonarCloud is flagging duplicated code in checkAnswers.js

brdunfield:
Although this does raise the question of whether the Ontario questions should show up on the checkAnswers page or not. I left them off for now, but they can easily be added.

routes/deductions/deductions.spec.js
```diff
@@ -95,7 +95,7 @@ describe('Test /deductions responses', () => {
             expect(response.headers.location).toEqual('/checkAnswers')
           } else {
             expect(response.headers.location).toEqual(
-              `${yesNoResponse.url}/amount?ref=checkAnswers`,
+              `${yesNoResponse.url.replace("/ontario", "")}/amount?ref=checkAnswers`,
```
pcraig3: lol

PR #398, opened 2019-12-09
Task/remove security questions 🚧💣

katedee:
[Trello Card](https://trello.com/c/7TPWbH74/469-s-remove-security-questions-remove-financial-info-page-remove-personalized-benefits-page-remove-tests)

This PR handles:
- removing security questions from the flow
- removing income + benefits information

Which includes removing all associated tests, schemas, views, controllers, etc.
Updated Cypress tests to reflect those changes.


cypress/integration/full_run_with_amounts.spec.js
```diff
@@ -320,7 +258,7 @@ describe('Full run through saying "yes" to everything', function() {
 
     //check some table data
     //until we have a more firm grasp on how we're shaping the total refund, i'm just checking benefits
-    checkTableRows(cy, getBenefitsBreakdownRows(this.user))
```
brdunfield: Is it still necessary to have two versions of the table / two util functions here?

katedee: Good point! I updated removing the non-lite version of checkBenefitsRows, and I modified checkTableRows to accept a string for css class name, so that we don't need a lite/non-lite version

cypress/integration/full_run_simple.spec.js
```diff
@@ -1,6 +1,5 @@
 const {
   checkTableRows,
-  checkTableRowsLite,
   getBenefitsBreakdownRowsLite,
```
brdunfield: Similar to my previous comment - I think this extends to element / function naming. If we no longer have a "Lite" mode, it's a bit confusing code-wise to have the word "Lite" sprinkled through our elements and functions.

cypress/utils.js
```diff
@@ -39,18 +29,6 @@ const getIncomeBreakdownRows = user => {
...
-      key: source.name,
-      value: `$${source.amount.toLocaleString('en-US')}`,
-    }
-  })
-
-  return benefitsRows
-}
-
-const getBenefitsBreakdownRowsLite = user => {
   const benefitsLiteKeys = Object.values(user.benefitsLite)
```
brdunfield: Just a couple last sneaky "Lite"s left!

routes/deductions/deductions.spec.js
```diff
@@ -9,7 +9,7 @@ describe('Test /deductions responses', () => {
 
   beforeEach(async () => {
     const testSession = session(app)
-    const getresp = await testSession.get('/financial/income')
+    const getresp = await testSession.get('/login/code')
```
pcraig3: Oh yeah, good catch.

pcraig3:
One last thing to remove is references to the `options` key since we don't use that anymore.

https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/utils/index.js#L428-L436

Other than that, merge away.

🔪

routes/deductions/deductions.spec.js
katedee: Less a catch, more a "how are 100+ tests failing I didn't even touch deductions" solution 😂

PR #402, opened 2019-12-11
Senior public transit tax

nmakuch:
- The work reflected in this PR is based off of the [Ontario seniors' public transit tax credit](https://trello.com/c/a4TtUEni), trello card

### Creating the Yes No page
- An opt in page has been created for the senior public transit tax
- Routes + Schemas + Error states have been added but could need to be updated when I talk to Brian tomorrow

Screen Shot 2019-12-10 at 23 21 47

### Creating the Amount page
- An amount page has been created for the senior public transit tax
- Routes + Schemas + Error states have been added but could need to be updated when I talk to Brian tomorrow

Screen Shot 2019-12-10 at 23 22 08


routes/confirmation/checkAnswers.js
```diff
@@ -87,6 +87,18 @@ const answerInfo = [
         urlPath: '/trillium/energy/cost/amount',
         displayIf: 'deductions.trilliumEnergyCostClaim',
       },
+      {
+        text: 'Did you use public transit in 2018?',
+        infoPath: ['deductions.seniorTransitClaim'],
+        urlPath: '/deductions/senior-public-transit',
+        displayIf: 'deductions.seniorTransitClaim',
+      },
```
katedee: I think we want this displayed regardless (so we want to show them saying No). So no need for the displayIf on the yes/no portion

routes/confirmation/checkAnswers.js
```diff
@@ -87,6 +87,18 @@ const answerInfo = [
...
+      {
+        text: 'Total amount spent on public transit',
+        infoPath: ['deductions.seniorTransitAmount'],
+        urlPath: '/deductions/senior-public-transit/amount',
+        displayIf: 'deductions.seniorTransitClaim',
```
katedee: but displayIf here is good👌

routes/confirmation/checkAnswers.js
nmakuch: I added some other stuff earlier today that I forgot to push up until now that probably needs to be looked over as well, and I'll confirm with Brian on this whether we want to always display it or not, I was just taking a best guess so you're probably right

routes/personal/personal.controller.js
```diff
@@ -76,12 +75,17 @@ const postConfirmMaritalStatus = (req, res, next) => {
     return res.redirect('/offramp/maritalStatus')
   }
 
+  // Just basing the difference in months or years is not a good enough check so we are checking the total delta days
+  if (getDateDelta(req.session.personal.dateOfBirth) === 23725 && confirmMaritalStatus === 'Yes') {
```
pcraig3: This will need to change every day, right?

pcraig3: Hey Nick! 

This works if someone finishes the marital status flow, but it doesn't guard against someone just straight up loading the `/deductions/senior-public-transit` URL in the browser. 

To solve this (ie, you shouldn't be able to see the `public-transit` urls at all depending on your date of birth), we should:

- point marital status pages at `public-transit` pages
- add middleware to all of the `public-transit` routes (GET and POST) to redirect to `trillium/rent` if we load this page but the DoB is wrong
- (since our tests will all fail, we can also include the `NODE_ENV !== 'production' ` in there or something) 
-

routes/deductions/deductions.controller.js
```diff
@@ -208,3 +282,15 @@ module.exports = function(app) {
...
+  return (req, res, next) => {
+    const dobInDays = getDateDelta(req.session.personal.dateOfBirth)
+    if (process.env.NODE_ENV === 'production') {
+      if (dobInDays <= SIXTYFIVE_YEARS_INDAYS) {
+        return res.redirect('/trillium/rent')
+      }
+    }
+    return next()
+  }
+}
```
pcraig3: We don't need the "return a function from a function" pattern here because we're not passing in anything to `seniorRedirect`.

We can rewrite this more simply as 

```
const seniorRedirect = (req, res, next) => {
    const dobInDays = getDateDelta(req.session.personal.dateOfBirth)
    if (process.env.NODE_ENV === 'production') {
      if (dobInDays <= SIXTYFIVE_YEARS_INDAYS) {
        return res.redirect('/trillium/rent')
      }
    }
    return next()
  }
```

And then we can call it on line 149, 155, etc as 

```
   seniorRedirect,
```

(without the brackets)

routes/deductions/deductions.controller.js
```diff
@@ -208,3 +282,15 @@ module.exports = function(app) {
     doRedirect,
   )
 }
+
+const seniorRedirect = () => {
+  return (req, res, next) => {
+    const dobInDays = getDateDelta(req.session.personal.dateOfBirth)
+    if (process.env.NODE_ENV === 'production') {
```
pcraig3: Also let's put a comment above this that explains why this check is here.

```// only skip this page in production. For testing and development, we want to be able to see the senior transit pages just like any other.```

PR #403, opened 2019-12-11
OTB Long-term care flow

brdunfield:
https://trello.com/c/IOGk9BZy/464-m-update-otb-long-term-care-flow

This PR spawns a couple associated issues that I've decided to split out into separate cards since they're a bit outside the scope of the original card.

Firstly, this PR introduces a new use-case which isn't covered by the current doYesNo middleware. I've created a card to address this issue here: https://trello.com/c/qzJUdDs8/516-m-refactor-doyesno-function-to-support-branching-decisions

Additionally, at this point deductions.spec is a bit of a mess from integrating all the new OTB flow pages / tests. Once all the OTB pages have been merged, I've created a card to clean up this file. More details here: https://trello.com/c/aTJDBuy6/515-s-refactor-tests-around-otb-flow

This flow is a bit more complex than some of the others that have gone in for OTB so far - more pages were added, and there's the new branching amounts flow (either the full amount or the room and board amount).


locales/en.json
```diff
@@ -1,4 +1,7 @@
 {
+  "45": "45",
+  "50": "50",
+  "4566": "4566",
```
pcraig3: Can we delete these numbers?

views/deductions/trillium-energy-cost.pug
```diff
@@ -11,6 +11,8 @@ block content
   form.cra-form(method='post')
     +radiosYesNo('trilliumEnergyCostClaim', 'In 2018, when you lived on an Ontario reserve, did you pay home energy costs, for example, to heat or cool your home, or use appliances like a stove or fridge?', trilliumEnergyCostClaim)
 
+    include ../_includes/principalResidence
```
pcraig3: Was this left out of a previous PR?

routes/deductions/deductions.controller.js
```diff
@@ -117,15 +120,52 @@ module.exports = function(app) {
...
+      'trilliumLongTermCareAmount',
+    ]),
+    doRedirect,
+  )
+
   app.get('/trillium/longTermCare/type', renderWithData('deductions/trillium-longTermCare-type'))
   app.post(
     '/trillium/longTermCare/type',
-    checkSchema(trilliumLongTermCareTypeSchema),
+    checkSchema(trilliumlongTermCareTypeSchema),
```
pcraig3: Weird that this shows up as an addition/deletion because these lines are identical, right?

brdunfield: The casing changed - there was some inconsistency with Long being spelled with a capital or lowercase. I made this one the same as the other schema.

views/deductions/trillium-energy-cost.pug
brdunfield: Ah, yes. I forgot to note it in my comment as I noticed it the other day and it slipped my mind today. It is supposed to be on this page per the content doc, but got missed in the PR for that page.

locales/en.json
brdunfield: Oops. I'll have to clean this up in a new PR. Maybe I'll bundle it in with the test cleanup issue?

PR #405, opened 2019-12-12
New page for when SIN and / or DoB do not match our records

brdunfield:
https://trello.com/c/UNX1akpi/468-new-page-for-bad-logins-ie-sin-and-or-dob-do-not-match-records

### DIFF
--- a/config/routes.config.js
+++ b/config/routes.config.js
@@ -5,6 +5,7 @@ const routes = [
   { path: '/offramp', editInfo: 'skip' },
   { path: '/login/sin' },
   { path: '/login/dateOfBirth' },
+  { path: '/login/error/doesNotMatch', editInfo: 'login.error'},
   { path: '/personal/name' },
   { path: '/offramp/name', editInfo: 'skip' },
   { path: '/personal/residence' },
--- a/config/routes.config.js
+++ b/config/routes.config.js
@@ -5,7 +5,7 @@ const routes = [
   { path: '/offramp', editInfo: 'skip' },
   { path: '/login/sin' },
   { path: '/login/dateOfBirth' },
-  { path: '/login/error/doesNotMatch', editInfo: 'login.error'},
+  { path: '/login/error/doesNotMatch', editInfo: 'skip'},
   { path: '/personal/name' },
   { path: '/offramp/name', editInfo: 'skip' },
   { path: '/personal/residence' },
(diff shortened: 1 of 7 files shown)


config/routes.config.js
```diff
@@ -5,6 +5,7 @@ const routes = [
   { path: '/offramp', editInfo: 'skip' },
   { path: '/login/sin' },
   { path: '/login/dateOfBirth' },
+  { path: '/login/error/doesNotMatch', editInfo: 'login.error'},
```
brdunfield: I added the editInfo here to a variable that doesn't exist - my reasoning is to ensure that this page doesn't show up when pressing "back" from the next page, but to allow it to stay here in the flow. If there's a better way to do this, happy to change it :)

pcraig3: We've done this in the past by using 'skip' (you can see above). It's doing what you are doing here, but it's easier to see which pages are skipped.

routes/login/login.session.spec.js
```diff
@@ -88,8 +88,8 @@ describe('Test /login SESSION responses', () => {
     expect(firstError.text()).toMatch('Please enter an access code')
   })
 
-  // TEST RETURN TO SIN PAGE IF SIN IS WRONG
-  test('it returns a 422 on /login/sin page when the SIN does not match the access code', async () => {
+  // TEST REDIRECT TO ERROR PAGE IF SIN IS WRONG
+  test('it returns a 200 on /login/error/doesNotMatch page when the SIN does not match the access code', async () => {
```
pcraig3: I think let's return the 422 if possible. Returning an error code is still the correct behaviour from an HTML point of view.

pcraig3: We can still create the errors key and get a 422 status from `renderWithData` like we did on the `/sin` page: https://github.com/cds-snc/cra-claim-tax-benefits/pull/405/files#diff-dd564f1ad0228a52ea372878f1aec040R25

(The error object itself wouldn't be used for anything.)

Also, we can remove the 'errorsKey' from the sin get render method.

brdunfield: Doing this doesn't appear to update the header location to the new URL - is that something we want to do? The URL stays on /dateOfBirth, and the tests are unable to check the header location.

brdunfield: Well, actually that was something else I tried. Doing the above still puts a big scary red box at the top of the page asking for errors to be corrected.

brdunfield:
Changed the route to a 'skip', but left the 200 responses for /doesNotMatch after speaking with Paul

pcraig3:
If you go through the app an put in the wrong sin, right dob, you the error the first time but then all your values are cleared.

If you do it the second time (same values), then they show up again. Is this just me or do you see it too?

PR #407, opened 2019-12-13
Refactor deductions.spec

brdunfield:
https://trello.com/c/aTJDBuy6/515-s-refactor-tests-around-otb-flow

Managed to basically half the size of the file while keeping all the same tests - should be much easier to read / write new tests with this format. Essentially added a "yesRedir" var to each yesNo URL so the tests know where the page should end up after a 'Yes' post.

There's one test that I will refactor once I address the doYesNo refactor, as the actual behaviour in the app isn't correct either and will be addressed when that issue is complete. I'll add a specific comment about it.


routes/deductions/deductions.spec.js
```diff
@@ -155,9 +104,9 @@ describe('Test /deductions responses', () => {
...
             .post(yesNoResponse.url)
             .use(withCSRF(cookie, csrfToken))
-            .send({ [yesNoResponse.key]: 'No', redirect: '/start' })
+            .send({ [yesNoResponse.key]: 'No', redirect: yesNoResponse.noRedir || '/start' })
           expect(response.statusCode).toBe(302)
-          expect(response.headers.location).toEqual('/start')
+          expect(response.headers.location).toEqual(yesNoResponse.noRedir || '/start')
         })
 
         test('it redirects to the checkAnswers when posting "No" and having come from the checkAnswers page', async () => {
```
brdunfield: This test will need a refactor due to the branching nature of the /longTermCare/type/cost page. Currently the page redirects to checkAnswers when a no is posted with the ref instead of the /longTermCare/type/amount page.

This will be addressed in https://trello.com/c/qzJUdDs8/516-m-refactor-doyesno-function-to-support-branching-decisions

routes/deductions/deductions.spec.js
```diff
@@ -14,114 +14,63 @@ describe('Test /deductions responses', () => {
...
-        const badValues = ['', null, false, 0, 'dinosaur', 'yes']
-        badValues.map(badValue => {
-          test(`it returns a 422 for a bad posted value: "${badValue}"`, async () => {
-            const response = await request(app)
-              .post(yesNoResponse.url)
-              .use(withCSRF(cookie, csrfToken))
-              .send({ [yesNoResponse.key]: badValue, redirect: '/start' })
-            expect(response.statusCode).toBe(422)
-          })
-        })
```
pcraig3: Are we losing these tests or do we have them elsewhere?

pcraig3: Nvm, just saw them.

pcraig3:
image

cut

pcraig3:
Hi @brdunfield, Nick just merged this one https://github.com/cds-snc/cra-claim-tax-benefits/pull/402, so now there's a conflict. We're not treating it any differently than other pages we have, so you should be able to add it into your array there.

brdunfield:
Yep! Merged! Also realized I had omitted the climate-action-incentive page, so it's been added as well.

PR #408, opened 2019-12-17
Add middleware to extend doYesNo for longTermCare branching

brdunfield:
https://trello.com/c/qzJUdDs8/516-s-refactor-doyesno-function-to-support-branching-decisions

The new OTB long term care flow has a branching flow where both 'Yes' and 'No' lead to amounts pages. 

As per @pcraig3 suggestion, instead of refactoring doYesNo itself, I added a unique middleware that clears the required session data on a 'Yes' response as well as a 'No' response for this post. This did involve some refactoring, namely to split out the code that clears the session so it could be re-used, and splitting out the code to find the next route so it could also be re-used (to avoid having to directly import routes.config in other files)

### DIFF
--- a/utils/index.js
+++ b/utils/index.js
@@ -185,31 +185,34 @@ const doYesNo = (claim, fields) => {
         return returnToCheckAnswers(req, res, true)
       }
 
-      const currentRoute = getRouteWithIndexByPath(req.path)
-      return res.redirect(defaultRoutes[currentRoute.index + 1].path)
+      return res.redirect(getNextRoute(req).path)
     }
 
     req.session.deductions[claim] = false
 
-    // fields is an array
-    fields.map(field => {
-      // if the field is truthy
-      if (req.session.deductions[field]) {
-        // "Amount" fields are zeroed -- others are nulled
-        if (field.endsWith('Amount')) {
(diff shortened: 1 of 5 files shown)


utils/index.js
```diff
@@ -415,8 +418,7 @@ const getPreviousRoute = (req, routes = defaultRoutes) => {
 }
 
 const returnToCheckAnswers = (req, res, claimYes = false) => {
-  const currentRoute = getRouteWithIndexByPath(req.path)
-  const nextRoute = defaultRoutes[currentRoute.index + 1]
+  const nextRoute = getNextRoute(req)
```
katedee: ✨

brdunfield:
I think I'll take care of it separately, I don't think this middleware should be affected.

utils/index.js
```diff
@@ -185,33 +185,36 @@ const doYesNo = (claim, fields) => {
         return returnToCheckAnswers(req, res, true)
       }
 
-      const currentRoute = getRouteWithIndexByPath(req.path)
-      return res.redirect(defaultRoutes[currentRoute.index + 1].path)
+      return res.redirect(getNextRoute(req).path)
```
pcraig3: Good refactor

PR #410, opened 2019-12-19
Added skip-link and removed autofocus attributes

brdunfield:
https://trello.com/c/70OAU6fD/524-s-add-skip-links-to-the-prototype

I didn't exactly copy the Gov.UK skip link, but put one together that I think is more similar to other skip links I've done with CDS in the past.

Noticing some weird cypress behaviour locally (I ran it a few times trying to debug a failed test, and it started working on its own..), so we'll see if the tests pass 🤞

### DIFF
--- a/public/scss/_base.scss
+++ b/public/scss/_base.scss
@@ -15,6 +15,32 @@ iframe,
   &:not([disabled]):focus { @include focus(); }
 }
 
+#skip-link-container {
+  width: 100%;
+  position: absolute;
+  z-index: 5;
+  text-align: center;
+  top: 10px;
+
+  #skip-link {
+    position: absolute;
+    width: 1px;
+    height: 1px;
+    overflow: hidden;
+    white-space: nowrap;
+    
+    &:focus {
+      position: static;
+      padding: 5px;
+      width: auto;
+      height: auto;
+      overflow: auto;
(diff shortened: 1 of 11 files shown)


public/scss/_base.scss
```diff
@@ -15,6 +15,32 @@ iframe,
   &:not([disabled]):focus { @include focus(); }
 }
 
+#skip-link-container {
+  width: 100%;
+  position: absolute;
+  z-index: 5;
+  text-align: center;
+  top: 10px;
```
pcraig3: We could use our generic spacing units for this, but I think just setting an hard-coded value is better (like you've done).

The skip link sits more or less outside of our spacing system as such, so this is a good way to go.

pcraig3:
Looks great, @brdunfield, thanks! 

One thing I noticed is that the anchor links in the URL don't go away between pages, although that's more to do with our redirects I think. 

For example, 
- if you click the skip link on the main page (`/start#content`), and then click "start now", the "#content" part of the URL goes away
- if you click the skip link on the `code` page (`/login/code#start`), and then put in a good code and hit "Continue", the next url is `/login/sin#code`. 

Seems like not the intended behaviour, but let's handle it outside of this PR.

PR #416, opened 2020-01-09
Eligibility criteria pre-screening

nmakuch:
- this work in this PR is related to the, [[Online channel] Eligibility criteria pages to be built](https://trello.com/c/lh6xiKCl) trello card
- this work needs to be merged for usability testing the week of Jan 20th

### i) Creating a Stepper mixin for these new pages as to better chunk out this section
- The stepper mixin will state the current step out of the total step

Screen Shot 2020-01-14 at 17 57 27

### ii) Creating the pages + routes for the Eligibility Criteria flow
- follows the flow outlined in this [google doc](https://docs.google.com/document/d/1MRRSB93wrwjMHWY1NTXBIpo-WMqNicwM68IVTMv8k94/edit)

eligibilityFlow

###  iii) Updating the login fixture for the cypress tests to include the new flow
- Adding the eligibility criteria to the routes page messed up the log in flow for the cypress tests
- The eligibility criteria pages were placed before the access code, sin and dob pages

### iv) Creating unit tests for these new pages + offramps
- followed the YesNo looping method from other .spec files
- copied this code over and created the data object to loop through

### v) Content updated on landing page
- Landing page has been changed to address the screening process

Screen Shot 2020-01-14 at 17 57 18

### vi) Conditional rendering on /taxable-income page and /offramp/taxable-income
- Rather than create 2 separate pages and mess routing up, a different amount is being rendered in the radio label whether the user is above or below 65
- The same is being applied to the offramp page

conditionalAge

### vii) Removed /personal/residence due to the new screening question
- screening question is replacing this page
- removed page, tests etc...


config/routes.config.js
```diff
@@ -1,6 +1,12 @@
 //list our routes in the flow order in the app
 const routes = [
   { path: '/start' },
```
pcraig3: Gotta update the "redirect" field in the `/start/index.pug` template.

routes/login/login.controller.js
```diff
@@ -133,3 +235,135 @@ const postLogin = async (req, res, next) => {
...
+
+const postTuition = (req, res, next) => {
+  const tuition = req.body.tuition
+
+  req.session.login.tuition = tuition
+
+  if (tuition === 'No') {
+    return res.redirect('/eligibility/income-sources')
+  }
+
```
katedee: for these ones where we run an `if NO` _and_ `if YES`, I assume one of the redirects we've put in the hidden redirect input, yes? So in this example, I think the NO url is in the redirect input, so we only need to check if tuition === YES and redirect. Otherwise, carry on with `next()` which should take to doRedirect.
Unless I've completely misunderstood, in which case lemme know!

pcraig3:
image

omg relax nick

api/user.json
```diff
@@ -1,7 +1,8 @@
 {
   "login": {
     "code": "A5G98S4K1",
-    "securityQuestion": null
+    "securityQuestion": null,
+    "ageYesNo": null
```
pcraig3: Interesting. This is _supposed_ to represent the data coming back from the API. We're using this here as a local session thing because we don't really have an API to worry about. 

Don't change it or anything — just making that observation for later.

cypress/integration/full_run_simple.spec.js
```diff
@@ -18,7 +18,7 @@ describe('Full run through saying "no" to everything', function() {
   // START PAGE
   it('successfully loads the home page', function() {
     cy.checkA11y()
-    cy.get('h1').should('contain', 'Claim Tax Benefits')
+    cy.get('h1').should('contain', 'File taxes to access benefits')
```
pcraig3: So this this the new name of our service? @brianhendrick1 @ClementineHahn @YedidaZalik

public/scss/_base.scss
```diff
@@ -141,6 +141,21 @@ li:not(:last-of-type) {
   @extend %page-container;
 }
 
+.stepper {
+  margin-bottom: 0;
+
+  p {
+  margin-top: $space-md;
+  margin-bottom: 5px;
```
pcraig3: Pretty sure we have a spacing var for `5`.

public/scss/_base.scss
```diff
@@ -141,6 +141,21 @@ li:not(:last-of-type) {
...
+  color: $stepper-grey-light;
+  }
+
+}
+
+.stepperH1 {
+  margin-top: 0;
+}
```
pcraig3: `+` is the "after" selector in CSS.

Unless you're doing something more elaborate, you can write this as:

```css
.stepper {
   ... CSS rules

  + h1 {
    margin-top: 0;
  }
}
```

public/scss/_variables.scss
```diff
@@ -13,6 +13,7 @@ $color-red-light: #f3e9e8;
 $color-purple: #7834bc;
 $color-grey-light: #cbcbcb;
 $color-grey-dark: #666666;
+$stepper-grey-light: #6f777b;
```
pcraig3: Visually, this is _very_ close to `666666`. Can we just use that one instead? (having lots of similar colours is the thing that variables are supposed to address).

pcraig3: Or we could change `$color-grey-dark` to `6f777b ` if that's a better solution.

routes/login/login.controller.js
```diff
@@ -133,3 +235,135 @@ const postLogin = async (req, res, next) => {
...
+  if (tuition === 'Yes') {
+    return res.redirect('/eligibility/tuition-claim')
+  }
+
+  next()
+}
```
pcraig3: To extend upon what @katedee said earlier, we also have a `doYesNo` function to abstract some of this stuff (ie, lots of repetition in here), but doYesNo is build on the assumption that "yes" === go to a new question in this tree, whereas "no" means "eff this and go to the next topic". 

For these questions, sometimes the "right" answer is "yes" (eg, do you live in Ont), other times it is "no" (ie, do you have kids), so that basic assumption is broken.

I'll bring it up with Brian, but it kind of looks like your "custom" approach to this makes sense for the time being.

views/login/eligibility-age.pug
```diff
@@ -0,0 +1,18 @@
+extends ../base
+
+block variables
+  -var title = __('Age 65 and above')
+  -var ageYesNo = hasData(data, 'login.ageYesNo') ? data.login.ageYesNo : ''
```
pcraig3: Looks like we're not actually using this variable down below (ie, it doesn't go anywhere).

stick it into the +radiosYesNo mixin as the `value=` so that the Yes/No is pre-checked when you reload the page for a question you've already answered.

views/login/eligibility-dependents-claim.pug
```diff
@@ -0,0 +1,21 @@
+extends ../base
+
+block variables
+  -var title = __('Tax credits for eligible dependents')
+  -var eligibleDependentsClaim = hasData(data, 'login.eligibleDependentsClaim') ? data.login.eligibleDependentsClaim : ''
+
+block content
+
+  +stepper('Screening questions', '4', '7')
```
pcraig3: Since this always says "screening questions", we can just put that in the macro.

views/_includes/stepperMixin.pug
```diff
@@ -0,0 +1,3 @@
+mixin stepper(category, currentStep, totalSteps)
+  .stepper
+    p #{__(category)} (#{'Step '} #{__(currentStep)} #{' of '} #{__(totalSteps)})
```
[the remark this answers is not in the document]
pcraig3: I noticed this below, but you can also just say `#{__('Screening questions')}` instead of `#{__(category)}`, since the category is always the same.

locales/en.json
```diff
@@ -1,4 +1,11 @@
 {
+  "1": "1",
+  "2": "2",
+  "3": "3",
+  "4": "4",
+  "5": "5",
```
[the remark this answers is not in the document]
katedee: maybe my view is out of date, but looks like we still have these numbers lurking around here in the translation

views/_includes/stepperMixin.pug
```diff
@@ -0,0 +1,3 @@
+mixin stepper(currentStep, totalSteps)
+  .stepper
+    p #{__('Screening questions')} (#{'Step '} #{__(currentStep)} #{' of '} #{__(totalSteps)})
```
katedee: So we still need to translate `Step`, just not the numbers. IT'll end up looking like:
`#{__('Step ')} #{currentStep} #{__(' of ')} #{totalSteps}`

nmakuch: I think this somehow made it back in when I rebased :s, I'll remove

views/login/eligibility-children.pug
```diff
@@ -0,0 +1,18 @@
...
+  -var children = hasData(data, 'login.children') ? data.login.children : ''
+
+block content
+
+  +stepper('3', '7')
+
+  h1 #{title}
+  
+  form.cra-form(method='post')
+    +radiosYesNo('children', 'Do you have a child or children under age 18?')
```
pcraig3: Don't forget to pass in the values for the rest of 'em. (here it would be `children `.)

pcraig3:
Hey @nmakuch, so I found one bug.

For the tuition or eligible dependents questions, if you say "yes", it takes you to a new question that you have to say "no" to (ie, you don't want to claim your tuition).

If you say "no" to tuition, you don't see the follow-up tuition question. 

But if you hit "go back", you always see the follow-up tuition question.

We have some logic in place to skip this stuff: the `editInfo` key, as you can see here: https://github.com/cds-snc/cra-claim-tax-benefits/blob/master/config/routes.config.js#L17

Add some of those to the routes and we should be all good.

nmakuch:
> Approved!
> 
> seal

😭

PR #418, opened 2020-01-13
changed alpha to beta in the header

charlesmorin:
Pretty much self explanatory.


charlesmorin:
Changed alpha to beta

PR #419, opened 2020-01-13
🧂Salt + Hash 🥔

katedee:
Based on some e-mail exchanges we ended up going with PBKDF2 at 100000 iterations. [Node has this built in](https://nodejs.org/api/crypto.html#crypto_crypto_pbkdf2sync_password_salt_iterations_keylen_digest), so we ended up using that.

A couple things to note:

### validate user
I've had to re-jig the way that validateUser works, because otherwise we would have to run through every entry and verify each hash. So I check by the code we already have, and then compare sin/dob

### static key vs random salt
Where I can, I’ve used a randomly generated salt. EXCEPT, for when initially checking and verifying an access code. I’ve used a static key value because otherwise we would have to query every row in the table, and compare the hash value. Fine when we have a small enough table, not so fine when we get a much larger user base.
Having a static key does present a minimal security risk to the access codes, but I think we’re fine considering everything else is a random salt.

Ideally we wouldn’t be keeping that static key (initialSalt) just written in some of our files right now, but sometimes we’re just running off the json file so  ¯\_(ツ)_/¯
I think maybe a better setup is having it be an ENV variable we don’t openly disclose on our repo, but open to other suggestions

### DIFF
--- a/api/user.json
+++ b/api/user.json
@@ -1,14 +1,14 @@
 {
   "login": {
-    "code": "A5G98S4K1",
+    "code": "d75535de98ecea315854491c8d036f8f$ada904924ee513b2d85d4ef775bbd1d835979dd8341047714da8d000ccec8adc",
     "securityQuestion": null
   },
   "personal": {
     "firstName": "Gabrielle",
     "lastName": "Roy",
     "confirmedName": null,
-    "sin": "540739869",
-    "dateOfBirth": "1977-09-09",
+    "sin": "3c983be4174c2711a44953311ee69792$cd18d36c21efd729f3ee8aab191912fb9467dbaba0552652cc6c451ac8a5031a",
+    "dateOfBirth": "5d3ac41f0d7b0387a651f2ed2b1d883a$fc2a8d33353897b1c6d723904911b91c6edb09283f7784e93ded0ef9dc659ba2",
     "address": {
       "line1": {
         "en": "303 Blake Blvd.",
--- a/api/user.json
+++ b/api/user.json
@@ -7,7 +7,7 @@
     "firstName": "Gabrielle",
     "lastName": "Roy",
     "confirmedName": null,
-    "sin": "540739869",
+    "sin": "3c983be4174c2711a44953311ee69792$cd18d36c21efd729f3ee8aab191912fb9467dbaba0552652cc6c451ac8a5031a",
     "dateOfBirth": "1977-09-09",
     "address": {
       "line1": {
--- a/api/user.json
+++ b/api/user.json
@@ -1,13 +1,13 @@
 {
   "login": {
-    "code": "d75535de98ecea315854491c8d036f8f$ada904924ee513b2d85d4ef775bbd1d835979dd8341047714da8d000ccec8adc",
+    "code": "A5G98S4K1",
     "securityQuestion": null
   },
   "personal": {
     "firstName": "Gabrielle",
     "lastName": "Roy",
     "confirmedName": null,
-    "sin": "3c983be4174c2711a44953311ee69792$cd18d36c21efd729f3ee8aab191912fb9467dbaba0552652cc6c451ac8a5031a",
+    "sin": "540739869",
     "dateOfBirth": "1977-09-09",
     "address": {
       "line1": {
--- a/api/user.json
+++ b/api/user.json
@@ -7,8 +7,8 @@
     "firstName": "Gabrielle",
     "lastName": "Roy",
     "confirmedName": null,
-    "sin": "3c983be4174c2711a44953311ee69792$cd18d36c21efd729f3ee8aab191912fb9467dbaba0552652cc6c451ac8a5031a",
-    "dateOfBirth": "5d3ac41f0d7b0387a651f2ed2b1d883a$fc2a8d33353897b1c6d723904911b91c6edb09283f7784e93ded0ef9dc659ba2",
+    "sin": "540739869",
+    "dateOfBirth": "1977-09-09",
     "address": {
       "line1": {
         "en": "303 Blake Blvd.",
--- a/db/db.json
+++ b/db/db.json
@@ -1,15 +1,15 @@
 [
   {
-    "code": "A5G98S4K1",
-    "sin": "540739869",
-    "dateOfBirth": "1977-09-09",
+    "code": "d75535de98ecea315854491c8d036f8f$ada904924ee513b2d85d4ef775bbd1d835979dd8341047714da8d000ccec8adc",
+    "sin": "3c983be4174c2711a44953311ee69792$cd18d36c21efd729f3ee8aab191912fb9467dbaba0552652cc6c451ac8a5031a",
+    "dob": "5d3ac41f0d7b0387a651f2ed2b1d883a$fc2a8d33353897b1c6d723904911b91c6edb09283f7784e93ded0ef9dc659ba2",
     "firstName": "Gabrielle",
     "locked": false
   },
   {
-    "code": "A5G98S4K2",
-    "sin": "536520422",
-    "dateOfBirth": "1987-05-05",
+    "code": "d75535de98ecea315854491c8d036f8f$04e748d8cf800dfbb0f1dd9e7cf57ec3204eaa291b271b3f0b0cb2ab4c6ad972",
+    "sin": "e14c9970c92a01b8b7c64164c2514c77$87c29bf7528aa0febbe3d83662f2d21720f0b0f074f8b9d7e0fcdbaf2647c9b8",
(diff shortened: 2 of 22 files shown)


katedee:
There's a small change around Date of Birth that needs to be made, but I think the core work can still be looked over and reviewed

utils/crypto.utils.js
```diff
@@ -0,0 +1,36 @@
...
+const config = {
+  initialSalt: 'd75535de98ecea315854491c8d036f8f',
+  iterations: 100000,
+  hashBytes: 32,
+  digest: 'sha512',
+}
+
+const hashArgs = [config.iterations, config.hashBytes, config.digest]
+
+const hashString = (stringToHash, useInitialSalt) => {
```
pcraig3: this one is good: `hashString('string')`
but this one is confusing: `hashString('string', true)`

could we use an options object for the last variable here? that way the intent is a bit clearer.

`const hashString = (stringToHash, { useInitialSalt } = {}) => {`

then to call it it looks like:

`hashString('string', {useInitialSalt: true})`

db/index.js
```diff
@@ -22,41 +26,47 @@ const useJson = (() => {
...
 
   const validateCode = async (code) => {
-    code = code.toUpperCase()
 
     if (useJson) {
-      return await jsonDB.find(user => user.code === code) || null
+      return await jsonDB.find(user => verifyHash(code.toUpperCase(), user.code, true)) || null
     }
 
+    code = hashString(code, true).catch(e => console.warn(e))
```
pcraig3: What are the scenarios that could cause an error, just an empty string at this point?

utils/crypto.utils.js
```diff
@@ -0,0 +1,36 @@
...
+  if (!stringToHash) {
+    return Promise.reject('you need to enter a string to hash')
+  }
+
+  const salt = useInitialSalt ? config.initialSalt : crypto.randomBytes(16).toString('hex')
+
+  const hash = crypto.pbkdf2Sync(stringToHash, salt, ...hashArgs).toString('hex')
+  return [salt, hash].join('$')
```
pcraig3: Nice, this looks really clean.

utils/crypto.utils.js
```diff
@@ -0,0 +1,36 @@
...
+}
+
+const verifyHash = (stringToVerify, original, useInitialSalt) => {
```
pcraig3: similar to ☝️ , let's do an object for the last one. 

`const verifyHash = (stringToVerify, original, { useInitialSalt }) => {`

Or we could make all of the params into object keys — whatever works:

`const verifyHash = ({ stringToVerify, original, useInitialSalt }) => {`

db/db.json
```diff
@@ -1,15 +1,15 @@
...
     "firstName": "Gabrielle",
     "locked": false
   },
   {
-    "code": "A5G98S4K2",
-    "sin": "536520422",
-    "dateOfBirth": "1987-05-05",
+    "code": "d75535de98ecea315854491c8d036f8f$04e748d8cf800dfbb0f1dd9e7cf57ec3204eaa291b271b3f0b0cb2ab4c6ad972",
+    "sin": "e14c9970c92a01b8b7c64164c2514c77$87c29bf7528aa0febbe3d83662f2d21720f0b0f074f8b9d7e0fcdbaf2647c9b8",
+    "dob": "21d3a8f03b9b135386839b597d18765a$438cab5c650846e0ff38ae91f09a1ae94bc461423caff37efb5f95a2545d3954",
```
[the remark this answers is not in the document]
katedee: yes, each salt is different and is randomly generated, which is a best practice for this.
I could change it to dateOfBirth, but our azure db is set up with `dob` (I based this off the init file we have for that), so we'd also have to talk to Charles to update that

db/index.js
```diff
@@ -22,41 +26,47 @@ const useJson = (() => {
...
-      return {"error": true}
+    if (!row) { return { "error" :  true } }
+
+    const incorrectInfo = [verifyHash(sin, row.sin), verifyHash(dateOfBirth, row.dob)].filter(v => v === false) 
+
+    if(incorrectInfo.length > 1) {
+      return { "error" :  true }
+    } else if (incorrectInfo.length) {
+      return null
     }
```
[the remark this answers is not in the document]
katedee: yeah, I tried following along when we send each use to what page and it became a bit confusing to map it all out. I think it was also easier to write before because we found a row by a matching SIn and DoB, but since that would require a ton of looping over each entry and verifying now, I pull in by the existing code we already have (so we only have to look up one row), and then check the SIN + DoB. I'd compare it against what we have now to sort of get what I'm saying 😅 
Happy to change this though.

utils/crypto.utils.js
```diff
@@ -0,0 +1,36 @@
+const crypto = require('crypto')
+
+const config = {
+  initialSalt: 'd75535de98ecea315854491c8d036f8f',
```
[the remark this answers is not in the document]
katedee: This is used for all codes. I explain in the PR description why we've used a static key for the codes

api/user.json
```diff
@@ -1,13 +1,13 @@
...
-    "code": "A5G98S4K1",
+    "code": "d75535de98ecea315854491c8d036f8f$ada904924ee513b2d85d4ef775bbd1d835979dd8341047714da8d000ccec8adc",
     "securityQuestion": null
   },
   "personal": {
     "firstName": "Gabrielle",
     "lastName": "Roy",
     "confirmedName": null,
-    "sin": "540739869",
+    "sin": "3c983be4174c2711a44953311ee69792$cd18d36c21efd729f3ee8aab191912fb9467dbaba0552652cc6c451ac8a5031a",
```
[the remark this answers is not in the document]
katedee: ah, okay. I assumed we'd still hash sensitive data like that, but can change it back to plain text if so

db/index.js
katedee: Yeah, basically. Which is not likely to happen, but just in case we've got there. We also reject the promise in hashSTring, and if we don't have a `catch` here then it's flagged as an unhandled promise rejection which is deprecated (and in the future will just exit our app).

db/db.json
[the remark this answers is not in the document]
pcraig3: Let's change to `dateOfBirth` and we can ask Charles to update the Azure DB.

db/index.js
```diff
@@ -22,41 +26,47 @@ const useJson = (() => {
...
 
   const validateCode = async (code) => {
-    code = code.toUpperCase()
 
     if (useJson) {
-      return await jsonDB.find(user => user.code === code) || null
+      return await jsonDB.find(user => verifyHash(code.toUpperCase(), user.code, {useInitialSalt: true})) || null
     }
 
+    code = hashString(code, {useInitialSalt: true}).catch(e => console.warn(e))
```
pcraig3: Does this need to be upper-cased as well?

db/index.js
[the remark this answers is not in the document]
pcraig3: Yeah, let's do another card for this but park it in the context of this PR. All good with this logic for the time being.

db/index.spec.js
```diff
@@ -1,9 +1,9 @@
 const DB = require('./index')
 
 const expectedRow = {
-  code: 'A5G98S4K1',
-  sin: '540739869',
-  dateOfBirth: '1977-09-09',
+  code: 'd75535de98ecea315854491c8d036f8f$ada904924ee513b2d85d4ef775bbd1d835979dd8341047714da8d000ccec8adc',
+  sin: '3c983be4174c2711a44953311ee69792$cd18d36c21efd729f3ee8aab191912fb9467dbaba0552652cc6c451ac8a5031a',
+  dob: '5d3ac41f0d7b0387a651f2ed2b1d883a$fc2a8d33353897b1c6d723904911b91c6edb09283f7784e93ded0ef9dc659ba2',
```
pcraig3: `dateOfBirth` here too.

utils/crypto.utils.js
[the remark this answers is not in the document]
pcraig3: Yup, you did.

db/index.js
katedee: no, because when we use the local db, it writes locally the hashed value into the session, and not the plaintext. So if we make it all caps the verify won't work

db/index.spec.js
katedee: hmm so it seems that camelCase [is possible in postgres, but can be a pain](https://stackoverflow.com/questions/11270753/underscores-or-camelcase-in-postgresql-identifiers-when-the-programming-languag). Snake case seems to be more standard, and is what's already happening in our cloud. But, open to changing it!

katedee: in which case, I'd also change to `first_name`

PR #431, opened 2020-01-17
change so we're not validating sin against date of birth

katedee:
for user testing, we don't want to validate the sin entered against the dob on record

to do so, I've take the check out of validateUser for the db, and some associated tests that were comparing the sin against the user info.

There were a couple to make sure that sin isn't null, but our schema handles a minimum length/entry on that front.

### DIFF
--- a/db/index.js
+++ b/db/index.js
@@ -1,8 +1,6 @@
 const { pool } = require('./config')
 const jsonDB = require('./db.json')
 
-const { cleanSIN } = require('../utils')
-
 const useJson = (() => { 
   if (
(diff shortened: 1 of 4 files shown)


pcraig3:
Hi @katedee, I'm asking @hillaryl if we still want the Luhn check as well. If not, then we remove that too and only check for (a) all digits, and (b) 9 characters.

pcraig3:
Hillary says we're all good to keep as-is. We want to run a test that's as close as possible to the live service (while obscuring the SIN), so we should throw SIN errors around the Luhn algorithm.

PR #432, opened 2020-01-20
Hide DB row names from the login controller

pcraig3:
The database rows aren't happy about uppercase letters, so it means we have to use snake_case row names (postgres convention) whereas the javascript variable names are all in camelCase (JS contvention).

We don't want different naming conventions in different parts of our code, so let's format the row names before they are returned. Pretty sure this works without making the code more complicated, but would like @katedee to have a look in particular.

This isn't strictly required, but I think it keeps the boundaries of the code clearer.

### DIFF
--- a/db/index.js
+++ b/db/index.js
@@ -5,36 +5,39 @@ const { cleanSIN } = require('../utils')
 
 const { verifyHash, hashString } = require('../utils/crypto.utils')
 
-
-const useJson = (() => { 
-  if (
-    process.env.USE_DB !== 'true' ||
-    process.env.NODE_ENV === 'test'
-  ) {
+const useJson = (() => {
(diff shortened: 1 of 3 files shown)


db/index.js
```diff
@@ -5,37 +5,55 @@ const { cleanSIN } = require('../utils')
...
+
+  const validateCode = async code => {
     if (useJson) {
-      return await jsonDB.find(user => verifyHash(code.toUpperCase(), user.code, {useInitialSalt: true})) || null
+      const row =
+        (await jsonDB.find(user =>
+          verifyHash(code.toUpperCase(), user.code, { useInitialSalt: true }),
+        )) || null
+
+      return _formatRow(row)
```
katedee: 👍

pcraig3:
Awesome, gonna merge. 👍

PR #448, opened 2020-01-27
Fix/check age gooder 👵

pcraig3:
## Check if someone was 65 in 2019, not if they are currently 65

When we are asking people's ages, we want to know if they were that age for a given tax year, not for the current year (in this case, 2019).

If someone turned 65 in January, it doesn't affect their taxes from the previous year. We want to know if they were 65 on December 31 of the previous year.

This means we can remove the `getDateDelta` function. Knowing someone's actual date on the day they're using the service isn't necessary, so removed the `getDateDelta` function entirely.

### DIFF
--- a/routes/deductions/deductions.controller.js
+++ b/routes/deductions/deductions.controller.js
@@ -8,7 +8,7 @@ const {
   renderWithData,
   checkErrors,
   postAmount,
-  getDateDelta,
+  is65,
 } = require('./../../utils')
 const {
   trilliumRentSchema,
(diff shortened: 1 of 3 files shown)


routes/deductions/deductions.controller.js
```diff
@@ -303,9 +301,9 @@ module.exports = function(app) {
 
 // Only skip this page in production. For testing and development, we want to be able to see the senior transit pages just like any other.
 const seniorRedirect = (req, res, next) => {
-  const dobInDays = getDateDelta(req.session.personal.dateOfBirth)
   if (process.env.NODE_ENV === 'production') {
-    if (dobInDays <= SIXTY_FIVE_YEARS_IN_DAYS) {
+    // if under 65, skip this page
+    if (!is65(req.session.personal.dateOfBirth)) {
       return res.redirect('/trillium/rent')
     }
```
nmakuch: nice 👍🏽

PR #459, opened 2020-01-28
remove dollar signs when posting amount values

katedee:
As detailed [in this card](https://trello.com/c/qTRdxcbZ/617-s-m-%F0%9F%95%B7adding-inputs-starting-with-dollar-signs-causes-a-bug), entering a dollar sign alongside the amount inputs causes `NaN` to display on CheckAnswers.

I've changed it so that we'll accept it on input, but strip it out in validation + formatting. This means navigating back to the page (or in case of errors being caught in schema validation) something like $100.30 will become 100.30 (or 100,30$ to  100,30 for french)

### DIFF
--- a/utils/index.spec.js
+++ b/utils/index.spec.js
@@ -123,7 +123,7 @@ describe('Test postAmount function', () => {
     },
     {
       input: '$1,000',
-      localse: 'en',
+      locale: 'en',
       expectedResult: '1000',
     },
     {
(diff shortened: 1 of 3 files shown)


utils/index.spec.js
```diff
@@ -121,6 +121,11 @@ describe('Test postAmount function', () => {
       locale: 'fr',
       expectedResult: '10341.28',
     },
+    {
+      input: '$1,000',
+      localse: 'en',
```
pcraig3: oops, it's `locale`

pcraig3: let's do a french one as well (unless we have it already in another test).

pcraig3:
Other than my previous comment, this should be good to go once you've quashed the LGTM alert ☝️

PR #464, opened 2020-01-31
Task/cleanup locales

katedee:
This is not an extensive locales cleanup (I'll still need to go through individually), but it still covers quite a bit.

I also added a utils files to the locales file, with some very basic functions to help us pick out lines that either are no longer in use, or need translation. It still has to be manually checked with a Search All in the editor, but helps hone it a bit. I've added them as scripts to package.json, but you're welcome to chuck it if you feel it's not useful.

The commands are:
- `missing-translations` => returns an array of all the lines in the FR locale file that currently are not translated, including some of the error messages which seem to often escape us!

- `missing-keys:en` => returns an array of all keys which are in EN file, but not in the FR file. This could mean it's no longer in use, or the french page has not been visited for this yet (thus lacking the key necessary for translation). We'll then have to manually check to see what the case is.

- `missing-keys:fr` => the same as above, except it returns keys in FR file that are not present in the EN locale file.


brdunfield:
I replaced the scripts from utils with the jsoncomb library, and added the "comb" script to use jsoncomb's normalize function. I ran this on the json files in locales, and then just for safety ran the cypress tests to make sure no changes showed up. @pcraig3 what do you think?

xml_output/taxfile-2018.xml
```diff
@@ -0,0 +1,94 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
pcraig3: We've updated the `.gitignore` so it doesn't ignore this file anymore, so let's delete it.

PR #465, opened 2020-02-03
Add checklist for our dependency update process

pcraig3:
I've described the real process I actually do so hopefully we don't have to say we do something else.

Inspired by the work @timarney did on the rescheduler here: https://github.com/cds-snc/ircc-rescheduler/blob/master/docs/upgrade.md

### DIFF
--- a/docs/UPDATING-DEPENDENCES.md
+++ b/docs/UPDATING-DEPENDENCES.md
@@ -18,7 +18,9 @@ We run this process at least once every 2-week period (ie, once a sprint), or as
 rm -rf package-lock.json ./node_modules
 ```
 
-4. Run the tests.
+4. Install new dependencies with `npm install`.
+
+5. Run the tests.
 
 ```
 npm test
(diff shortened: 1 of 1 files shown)


docs/UPDATING-DEPENDENCES.md
```diff
@@ -0,0 +1,40 @@
+# Updating dependencies
+
+We use open-source npm dependencies to build our application. These dependencies are (mostly) under active development and new versions are pushed fairly regularly.
+
+As many dependency updates fix small bugs or ensure compatibility with new platforms, we want to be keeping up with them. Letting dependencies fall behind several major versions is generally a bad scene.
```
damcom333: For sure a bad scene.

docs/UPDATING-DEPENDENCES.md
```diff
@@ -0,0 +1,40 @@
...
+
+```
+npm test
+npm run lint
+npm run cypress:cli
+```
+
+5. If the tests _fail_, continue to step 6. If the tests _pass_, continue to step 7.
+
+6. Revert to the previous dependency versions (eg, `git checkout -- .`) and run `npm install` to re-download them. Run the tests with the original dependencies to ensure they do actually pass. If they do, you can be sure that your updates broke the tests and not some other random issue.
```
damcom333: 👍

sboisvert:
@damcom333 What are your thoughts on this being the starting point for our defaults for the Tech best practices guide?

damcom333:
@sboisvert It's good. It does need extra instructions on how to release the update all the way to anything live on the internet.

sboisvert:
> all the way to anything live on the internet.

Would that be part of the CI/CD pipeline documentation or do you see it as something separate for updates?

damcom333:
It's most likely a very specific thing, every time. The last mile to release anything is most always unique around these parts. So yes to a pipeline that goes all the way. But also yes to the laborious process that gets it to that live instance.

sboisvert:
>The last mile to release anything is most always unique around these parts. 

By unique do you mean unique to each app, or unique to each push to production?

pcraig3:
@damcom333 for this particular app, the last mile is:
- (Submit pull request) 
- Get peer reviewed → Merge PR
- You're done! Merging the pull request will automatically deploy the service

Is that worth putting in here?

damcom333:
I think it's worth mentioning that CD will carry the patch into live system. But yes, thanks for clarifying that CD does finish the job.

docs/UPDATING-DEPENDENCES.md
```diff
@@ -0,0 +1,40 @@
...
+
+   - Return to step 1.
+   - When you arrive at step 2, only apply the **minor**, and **patch** upgrades.
+   - If any of the tests fail on step 4, revert again but this time apply only the **patch** upgrades.
+   - If the tests continue to fail on the patch upgrades, debug the problem(s) until the tests pass and the application works as expected.
+
+7. Boot up the application locally and make sure things look as expected (the cypress tests (`npm run cypress:cli`) cover most of the pages, so this is just a sense check).
+
+8. Submit a pull request with your updates.
```
damcom333: Be explicit here about how the change gets applied to production. So yes, the CD takes care of it.

docs/UPDATING-DEPENDENCES.md
```diff
@@ -0,0 +1,64 @@
...
+
+2. Apply all updates — patch, minor, and major updates — by updating the version numbers in `package.json` to match the version in the `LATEST` column returned by `npm`.
+
+3. Delete the `package-lock.json` file and the `./node_modules` folder. This will ensure we get the latest nested dependencies as well.
+
+```
+rm -rf package-lock.json ./node_modules
+```
+
+4. Run the tests.
```
brdunfield: I think we need to run an npm install to get the node_modules folder back before running the tests.

PR #470, opened 2020-02-06
Fix/Remove locales lines we don't need

pcraig3:
Went through the locales files again and there were a lot of lines in there that we aren't using. It's possible I took out more than I should of (ie, a few lines might show up now and again), but better that than leaving stuff in forever that we don't need.

Can't think of a good automated way to do this, but we don't really need an automated way to do it. Just cleaning them up every now and then is good.

**Also**

Removed the `/offramp` url which we aren't actually using any longer.


routes/offramp/offramp.spec.js
```diff
@@ -3,22 +3,21 @@ const app = require('../../app.js')
...
-    '/offramp/securityQuestion',
+    '/offramp/foreign-income',
+    '/offramp/income-sources',
+    '/offramp/taxable-income',
   ]
 
   urls.map(url => {
     test(`it returns a 200 response for ${url}`, async () => {
-      const response = await request(app).get('/offramp')
+      const response = await request(app).get(url)
```
brdunfield: Good catch. 👀

pcraig3: Yeah, these tests would all pass all the time because we were testing the same page everytime. 🤔

Issue #468, opened 2020-02-06
Typography nerd stuff 🤓


pcraig3:
# Typography nerd stuff 🤓

Okay, let's talk about typography for a bit.

Very early on we started building this site with a “we need a demo in 3 weeks” timeline.

As a starting basis, I grabbed the Canada.ca theme from CRA and we build out the first bunch of pages that way. Here is an example Canada.ca page vs our first-ever home page.

| CRA website ([E-services](https://www.canada.ca/en/revenue-agency/services/e-services/e-services-businesses.html)) | CTB homepage  |
|--------|-------|
| <img width="1469" alt="Screen Shot 2020-02-05 at 6 22 45 PM" src="https://user-images.githubusercontent.com/2454380/73892307-97cd1f80-4844-11ea-99fe-36aa4265f899.png">  | <img width="1469" alt="Screen Shot 2020-02-05 at 6 21 52 PM" src="https://user-images.githubusercontent.com/2454380/73892260-6c4a3500-4844-11ea-9b5d-3bc404b0b3a9.png">     |

After the initial setup, there were basically 2 rounds of pull request specifically about on our site-wide typography.

1. [Fixed some layout bugs to normalize margins and bottom spacing](https://github.com/cds-snc/cra-claim-tax-benefits/pull/13) (click for pics)
2. [Adjusted line height and shrunk text sizes on mobile](https://github.com/cds-snc/cra-claim-tax-benefits/pull/102) (also click for pics)

Obviously we've done a lot of one-off CSS adjustments and we've added new components from time to time as well, but those two pull requests above are the main "site-wide typography rationalizations" we've done so far.

## No clear guidelines in the Canada.ca theme

The Canada.ca theme doesn’t really have an opinion on font measure (ie, 'measure' == line length). On a large screen, a 2/3rds column is roughly the right line length, but if we use CRA as an example, there's full-width lines all over the place. 

| [NETFILE - Overview](https://www.canada.ca/en/revenue-agency/services/e-services/e-services-individuals/netfile-overview.html) | [E-services](https://www.canada.ca/en/revenue-agency/services/e-services.html) | [E-services for businesses](https://www.canada.ca/en/revenue-agency/services/e-services/e-services-businesses.html) |
|-------|--------|-------|
|  ✅2/3rds: ~70-80 character lines   |  ❌ full-width: > 110 character lines    |   ❌ full-width: > 110 character lines    |
[...]

PR #471, opened 2020-02-06
Typography fixes

brdunfield:
https://github.com/cds-snc/cra-claim-tax-benefits/issues/468

Addresses the 3 major issues from the linked issue:
1. Question text / hint text length has been made longer
2. Full width paragraphs have been shortened to match the start and confirmation pages
3. Titles will wrap to a second line earlier, reducing the max-width of the title.

### DIFF
--- a/public/scss/_forms.scss
+++ b/public/scss/_forms.scss
@@ -35,6 +35,13 @@ form.cra-form {
 
   @include md {
     max-width: 80%;
+
+    .w-3-4 {
+      width: 50%;
+    }
+    .w-1-2 {
+      width: 30%;
+    }
   }
 
   > div:not(:last-of-type) {
@@ -176,6 +183,7 @@ form.cra-form {
     display: flex;
     justify-content: space-between;
     flex-direction: column;
+    max-width: 450px;
 
     .buttons--left,
     .buttons--right {
--- a/public/scss/_forms.scss
+++ b/public/scss/_forms.scss
@@ -33,6 +33,10 @@ form.cra-form {
     }
   }
 
+  @include md {
+    max-width: 80%;
+  }
+
   > div:not(:last-of-type) {
     margin-bottom: $space-xl;
   }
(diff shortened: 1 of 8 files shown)


pcraig3:
@brdunfield, almost there. Sorry for being really picky but 2 things 🙏 

1. We can skip the header padding on small screens since it's only on large screens we see the super long line lengths.

| before | after |
|--------|-------|
|  header text has no padding (looks okay on a small screen)   |  header text has a padding (let's skip it) |
| Screen Shot 2020-02-07 at 3 57 38 PM   | Screen Shot 2020-02-07 at 3 57 36 PM   |

2. Inputs: let's keep 'em the same max width as they are on the current "live" service. (I know when you shrink the screen they get smaller — that's okay)

Here's a gif that shows them going back and forth. (The longer inputs are the new CSS.)

inputs

So yeah, let's max them out at the current "live" size.

public/scss/_forms.scss
```diff
@@ -37,10 +37,10 @@ form.cra-form {
     max-width: 80%;
 
     .w-3-4 {
-      width: 50%;
+      width: 338px;
     }
     .w-1-2 {
-      width: 30%;
+      width: 225px;
```
pcraig3: Ah, ha. So we do want to cap the width but doing this changes the intent of the classname from a relative sized-element to an absolute one.

If we set both `width` and `max-width` properties, we can allow the field to keep its relative sizing but also cap its length at some rough ~number of character size.

Here's an example: https://jsbin.com/geqohusomi/edit?html,output

In the example, the width of the input is 50% unless it exceeds 30 characters (type in 30 characters to check), at which point it is capped.

We pretty much have 3 kinds of text input fields at this point:
- login code / SIN (sized around 30 characters)
- date of birth: very small inputs
- amounts of money: probably max out around 15 characters (although that doesn't include the dollar sign)

The existing app is a pretty good reference for these sizes (obviously they are bigger but we don't want to make the inputs too short), so play around setting 'ch' values as `max-widths` so that the inputs look pretty similar to currently but can still grow and shrink.

Almost there!

t

PR #474, opened 2020-02-10
[Very meta] Add code review guidelines

pcraig3:
@brdunfield, sorry to make you the first-ever person to be held to this Very High standard but someone's gotta be.

***

I went over a bunch of different sources for this.
I tried to split the difference and come up with something that is useful for reviewers to read, both in terms of the tone of reviews and in the more technical terms of what we should look for.

Here are my sources:

- https://github.com/alphagov/styleguides/blob/master/pull-requests.md
- https://github.com/cds-snc/.github/blob/master/CODE_OF_CONDUCT.md
- https://github.com/alphagov/styleguides/blob/master/git.md
- https://github.com/thoughtbot/guides/tree/master/code-review
- https://www.evoketechnologies.com/blog/code-review-checklist-perform-effective-code-reviews/
- https://www.codementor.io/blog/code-review-checklist-76q7ovkaqj
- https://www.michaelagreiler.com/code-review-checklist/

### DIFF
--- a/docs/CODE-REVIEW-GUIDELINES.md
+++ b/docs/CODE-REVIEW-GUIDELINES.md
@@ -2,14 +2,14 @@
 
 While solipsism is an interesting philosophical tributary, it’s not a very good way to write long-term maintained software. In general we work in teams at CDS, so we require mandatory code reviews before merging any Pull Request.
 
-Just like all the fun and cool code we love to write often comes down to subjective, individual decisions, the practice of reviewing code is pretty subjective, given that different individuals will do different reviews.
+Just like how all the fun and cool code we love to write often comes down to subjective, individual decisions, the practice of reviewing code is pretty subjective as well — different individuals will do different reviews.
 
-Here are a list of considerations to keep in mind when doing a review: firstly focusing on principles to keep in mind for reviewers, and then leading into technical considerations.
+While recognizing this, here are a list of considerations to keep in mind when doing a review: firstly focusing on principles to keep in mind for reviewers and reviewees, and then leading into technical considerations.
 
 ## Principles for reviewers
 
 - [Be nice](https://qz.com/work/625870/after-years-of-intensive-analysis-google-discovers-the-key-to-good-teamwork-is-being-nice/)
-  - If you're not sure how someone wants their code reviewed or how they prefer to receive feedback, talk to them beforehand
+  - If you’re not sure how someone wants their code reviewed or how they prefer to receive feedback, talk to them beforehand
   - read the CDS [Code of Conduct](https://github.com/cds-snc/.github/blob/master/CODE_OF_CONDUCT.md)
 - Point out when something is particularly good or clever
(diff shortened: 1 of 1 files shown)


docs/CODE-REVIEW-GUIDELINES.md
```diff
@@ -0,0 +1,85 @@
...
+  - Make sure to test the working code yourself
+- Is all the code easily understood?
+  - Which parts are confusing to you and why?
+  - Multiple (nested) if/else blocks are code smells for complex logic
+  - If you are taking a lot of time to understand the code, perhaps it needs comments or refactoring
+- Can the code be simplified?
+  - Remove unused parameters or "generic" code that is only used once
+  - Don't solve problems you don't have
+- Does similar functionality already exist in the codebase?
+  - If so, why isn’t this functionality reused? (Note: it's okay to not reuse things — but it's worth asking the question.)
```
brdunfield: I think this touches on an important point - which is that as a review-ee, if someone reviewing your code is asking a question about something, that doesn't mean you've done it wrong. It's perfectly ok to justify your choices and stand by them.

Which is leading me to wonder if maybe a short section on review-ee guidelines might be a good idea?

docs/CODE-REVIEW-GUIDELINES.md
```diff
@@ -0,0 +1,85 @@
...
+- Are tests checking things at the right level?
+  - Often, unit tests aren't suitable for complex flows
+- Do the existing set of tests pass?
+  - Tests should be running as part of a continuous integration workflow
+
+### Documentation
+
+- Does this change require an update to the README or configuration files?
+- Is the reason for the change outlined in the commit message?
+  - [A good commit message](https://github.com/alphagov/styleguides/blob/master/git.md) explains what changed and _why_ it was done this way
```
brdunfield: This would be something I'd personally need to work on, but agree with. VS Code for some reason tries to get scary with you if your commit message is over 50 characters in a single line (and I only just realized that this is only talking about a single line and not the message as a whole), which has influenced me towards brevity in the commit message with longer PR descriptions.

Having reasons be documented in the commit message is definitely worth doing.

docs/CODE-REVIEW-GUIDELINES.md
pcraig3: Yeah, not a bad idea. I'll add some.

docs/CODE-REVIEW-GUIDELINES.md
pcraig3: Writing good commits is important! Basically what you should be thinking is that if someone comes along in a year to figure out why a change was made, the commit should explain it. (Some changes are very obvious, but not all.)

But I agree that unless you know pretty well how VIM works, the tools to do it are often pretty bad.

docs/CODE-REVIEW-GUIDELINES.md
pcraig3: Updated!

PR #475, opened 2020-02-11
Voter pages content restructure

brdunfield:
https://trello.com/c/K9Fpieco/656-m-update-pages-about-voting

- Removed old confirmation page
- Replaced with new 'citizen' and 'register' pages as per the content docs

This was pretty similar to the work we did on adding / expanding the Trillium pages. Hopefully I remembered all the details.


locales/fr.json
```diff
@@ -307,8 +311,6 @@
   "errors.login.trusteeLastName": "Veuillez écrire le nom de famille de votre syndic de faillite",
   "errors.maritalStatus.maritalStatus": "Vous devez choisir l’une des options",
   "errors.review": "Vous devez accepter de produire votre déclaration de revenus. Cochez la case si vous acceptez.",
-  "errors.voterCitizen": "You must be a Canadian citizen to register to vote.",
-  "errors.voterConsent": "You must allow CRA to share information with Elections Canada to register to vote",
```
pcraig3: 👏  Nice!

routes/vote/vote.controller.js
```diff
@@ -30,10 +42,10 @@ const postOptIn = (req, res, next) => {
 
   // if yes, go to second page of vote
   if (confirmOptIn === 'Yes') {
-    return res.redirect('/vote/confirmRegistration')
+    return res.redirect('/vote/citizen')
   }
   // if no, go to confirmation
-  req.session.vote.voterCitizen = null
+  req.session.vote.citizen= null
   req.session.vote.voterConsent = null
```
pcraig3: Pretty sure this is just `register` now.

routes/vote/vote.controller.js
```diff
@@ -42,9 +54,15 @@ const postOptIn = (req, res, next) => {
...
-const postConfirmRegistration = (req, res, next) => {
-  req.session.vote.voterCitizen = req.body.voterCitizen == 'voterCitizen' ? 'Yes' : 'No'
-  req.session.vote.voterConsent = req.body.voterConsent == 'voterConsent' ? 'Yes' : 'No'
+const postCitizen = (req, res, next) => {
+  const citizen = req.body.citizen
+  req.session.vote.citizen = citizen
 
+  if (citizen === 'Yes') {
+    return res.redirect('/vote/register')
+  }
```
pcraig3: We should null our the `register` value  here too if the answer is no.

views/vote/citizen.pug
```diff
@@ -0,0 +1,20 @@
+extends ../base
+include ../_includes/banner
+
+block variables
+  -var title = __('Register to vote - citizenship')
+  -var citizen = !hasData(data, 'vote.citizen') ? '' : data.vote.citizenship
```
pcraig3: Ah ha, pretty sure this should be "data.vote.citizenship".

routes/vote/vote.controller.js
```diff
@@ -42,9 +54,15 @@ const postOptIn = (req, res, next) => {
...
 
-const postConfirmRegistration = (req, res, next) => {
-  req.session.vote.voterCitizen = req.body.voterCitizen == 'voterCitizen' ? 'Yes' : 'No'
-  req.session.vote.voterConsent = req.body.voterConsent == 'voterConsent' ? 'Yes' : 'No'
+const postCitizen = (req, res, next) => {
+  const citizen = req.body.citizen
+  req.session.vote.citizen = citizen
 
+  if (citizen === 'Yes') {
+    return res.redirect('/vote/register')
```
pcraig3: Small comment (don't know if it matters but what do you think):

- If you come back here from checkAnswers and then click "yes", you have to fill out the rest of the flow (see example)


vote

It's not so many pages so it's not a big deal, but technically we should preserve the "checkAnswers" value.

pcraig3:
Hi @YedidaZalik, I have a small content revision to you.

Someone who says they are interested in voting but then says "No" to the next page (ie, if they are a Canadian Citizen / If they want to send their info), sees this on the `/checkAnswers` page.

Screen Shot 2020-02-12 at 3 58 45 PM

The short title says "Registered to vote: Yes", which is incorrect. They indicated interest in registering but they didn't make it, so the title should say something else.

views/vote/citizen.pug
brdunfield: Actually the second part should be data.vote.citizen! Originally I started with a mix of citizenship and citizen, and eventually picked one to unify. This must be a leftover.

routes/vote/vote.controller.js
brdunfield: Agreed, I'll fix this.

PR #476, opened 2020-02-13
Fix some values not being prefilled

brdunfield:
https://trello.com/c/ISdkyjLO/662-xs-%F0%9F%90%9C-values-arent-being-pre-filled-all-the-time

- The confirm income box was a missing variable
- The amounts for long term care were being intentionally cleared to make sure if the selection had changed that the old data didn't stick around. I made it a little smarter so it only clears if the selection actually has been changed.

### DIFF
--- a/routes/deductions/deductions.controller.js
+++ b/routes/deductions/deductions.controller.js
@@ -248,7 +248,6 @@ module.exports = function(app) {
       // if there is already data and this selection is the opposite of what's saved,
       // we need to clear the data
       if (req.session.deductions[claim] && req.session.deductions[claim] != (req.body[claim] === 'Yes' ? true : false)) {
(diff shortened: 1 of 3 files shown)


brdunfield:
I seem to have run into the same bug that @katedee  was having with Cypress awhile back. When running through the second fixture, the box is already checked from the previous session, so when the fixture clicks it, it unchecks and errors out.

The simplest solution is to change the fixture so it doesn't click the box in this fixture, but I don't LOVE doing that since it's not technically correct.. I'm going to look into cypress a bit and see if I can find a way to restore the session between fixtures, maybe as another PR? @pcraig3 Thoughts?

brdunfield:
Merged in the fix from #478

PR #529, opened 2020-02-26
Skip adding index number to first radio input

katedee:
This PR changes it so that we skip adding the index number to the first radio input, in a list of radio buttons. This is so that our errorlist can then link to the first radio input, for better accessibility. Cypress tests updated as well

There shouldn't be conflict in the IDs, but if we feel we want to be extra careful, I can add a string at the end while generating the IDs? But looking as it is now, this shouldn't be happening so I'd like to avoid tacking on things


katedee:
Since our heroku deploys are having A Time™ linking to our PRs at the moment, here's the link to this deployment:
https://cra-claim-ta-fix-radio-mhsfmkk.herokuapp.com/start

PR #542, opened 2020-03-02
Remove the if65() function

### LINKED ISSUE #537: https://claim-tax-benefits.azurewebsites.net/confirm-income
This page should give a different amount for people who are 65 or older, but when I tried using that age  it was the 12K amount. For 65 and older, 

Please replace first line with:
If your income from all sources is $19,564 or more:

Please change last line to:
Check this box to confirm you income from all sources was less than $19,564 in 2019

Thanks!

katedee:
as now we log in user.json if they answer yes or no to being 65 and above
This isn't necessary though, and this PR is very take-it-or-leave-it

The benefit to removing it is just having that little bit less of code to maintain
I've marked it as "resolving" an issue, but in fact, the issue is resolved (full context in a comment on the issue)

- Resolves #537 

image


katedee:
Closing because there's good reason not to go with this as outlined here:
https://github.com/cds-snc/cra-claim-tax-benefits/issues/537#issuecomment-593487860

PR #569, opened 2020-09-24
Update dependencies + remove sonarqube

pcraig3:
This PR

- Updates all the app dependencies other than cypress (the cypress tests crash otherwise)
- Removes sonarqube from the equation

We don't really need sonarqube in the context of our prototype.


pcraig3:
Gonna merge, pretty sure we're good here.

